PR #47, opened 2022-08-23
[MI-2072] Add testcases for configuration

ayusht2810:


raghavaggarwal2308 (approved):

server/config/config_test.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+func TestIsValid(t *testing.T) {
+	for _, testCase := range []struct {
+		description string
+		config      *Configuration
+		errMsg      string
+	}{
+		{
+			description: "valid configuration",
```
avas27JTG, line 17:
```suggestion
			description: "valid configuration",
```

we can either write a proper statement or sentence like
```
with valid configuration
with invalid configuration where AzureDevopsAPIBaseURL is empty
```

or can write as sort description like 
```
configuration: valid,
configuration: empty AzureDevopsAPIBaseURL
```

fix for all such places

server/config/config_test.go
```diff
@@ -0,0 +1,152 @@
...
+		},
+		{
+			description: "invalid configuration with AzureDevopsAPIBaseURL empty",
+			config: &Configuration{
+				AzureDevopsAPIBaseURL:        "",
+				AzureDevopsOAuthAppID:        "mockAzureDevopsOAuthAppID",
+				AzureDevopsOAuthClientSecret: "mockAzureDevopsOAuthClientSecret",
+				EncryptionSecret:             "mockEncryptionSecret",
+			},
+			errMsg: "azure devops API base URL should not be empty",
```
avas27JTG, line 34: create constant for this in message.go and use both here as well as in config file

avas27JTG (approved):

PR #56, opened 2022-08-24
[MI-2087] Add missing serializers, validations and middleware

ayusht2810:


server/plugin/api.go
```diff
@@ -91,11 +90,11 @@ func (p *Plugin) handleCreateTask(w http.ResponseWriter, r *http.Request) {
...
-	decoder := json.NewDecoder(r.Body)
-	if err := decoder.Decode(&body); err != nil {
+
+	body, err := serializers.LinkPayloadFromJSON(r.Body)
+	if err != nil {
 		p.API.LogError(constants.ErrorDecodingBody, "Error", err.Error())
-		p.handleError(w, r, &serializers.Error{Code: http.StatusInternalServerError, Message: err.Error()})
+		p.handleError(w, r, &serializers.Error{Code: http.StatusBadRequest, Message: err.Error()})
 		return
 	}
```
Kshitij-Katiyar, line 100: at line 102-103 dont create a variable use that directly

server/plugin/api.go
```diff
@@ -176,14 +175,18 @@ func (p *Plugin) handleGetAllLinkedProjects(w http.ResponseWriter, r *http.Reque
...
-	decoder := json.NewDecoder(r.Body)
-	if err := decoder.Decode(&project); err != nil {
+	project, err := serializers.ProjectPayloadFromJSON(r.Body)
+	if err != nil {
 		p.API.LogError(constants.ErrorDecodingBody, "Error", err.Error())
 		p.handleError(w, r, &serializers.Error{Code: http.StatusBadRequest, Message: err.Error()})
 		return
 	}
 
+	if validationErr := project.IsValid(); validationErr != "" {
```
Kshitij-Katiyar, line 185: why return string rather than error you are returning error in every other isValid function

server/serializers/project.go
```diff
@@ -1,8 +1,37 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"io"
+
+	"github.com/Brightscout/mattermost-plugin-azure-devops/server/constants"
+)
+
 type ProjectDetails struct {
 	MattermostUserID string `json:"mattermostUserID"`
```
raghavaggarwal2308, line 12: Is mattermostuserID optional?

raghavaggarwal2308 (changes requested):

server/serializers/project.go
ayusht2810: yes, as we extract it from the header whenever we require it

raghavaggarwal2308: okay

raghavaggarwal2308 (approved):

Kshitij-Katiyar (approved):

avas27JTG (approved):

PR #58, opened 2022-08-25
[MI-2092] Add success modal to the modals

ayusht2810:
image
image

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -43,7 +43,7 @@
                 "key": "EncryptionSecret",
                 "display_name": "Encryption Secret:",
                 "type": "generated",
-                "help_text": "The secret key is used to encrypt and decrypt the OAuth token.\nRegenerating the secret key will cause the loss of data."
+                "help_text": "The secret key is used to encrypt and decrypt the OAuth token.\nRegenerating the secret will require all users to re-connect their accounts to Azure DevOps."
             }
         ]
     }
--- a/plugin.json
+++ b/plugin.json
@@ -43,7 +43,7 @@
                 "key": "EncryptionSecret",
                 "display_name": "Encryption Secret:",
                 "type": "generated",
-                "help_text": "The secret key is used to encrypt and decrypt the OAuth token.\nRegenerating the secret will require all users to re-connect their accounts to Azure DevOps."
+                "help_text": "The secret key used to encrypt and decrypt the OAuth token.\nRegenerating the secret will require all users to re-connect their accounts to Azure DevOps."
             }
         ]
     }
--- a/webapp/src/components/resultPanel/index.tsx
+++ b/webapp/src/components/resultPanel/index.tsx
@@ -28,7 +28,7 @@ const ResultPanel = forwardRef<HTMLDivElement, ResultPanelProps>(({
     >
         <>
             {iconClass ? (
-                <i className={`fa result-panel-icon ${iconClass ?? 'fa-check-circle-o'}`}/>
+                <i className={`fa result-panel-icon ${iconClass || 'fa-check-circle-o'}`}/>
             ) : (
                 <SVGWrapper
                     className='result-panel-icon'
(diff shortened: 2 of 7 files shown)


webapp/src/components/resultPanel/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+                    {SVGIcons.check}
+                </SVGWrapper>
+            )}
+            <h3 className='result-panel-text'>{header ?? 'Add new'}</h3>
+            {onPrimaryBtnClick && (
+                <button
+                    className='btn btn-primary'
+                    onClick={onPrimaryBtnClick}
+                >
+                    {primaryBtnText ?? 'Create new'}
```
raghavaggarwal2308, line 48: Is the value of 'primaryBtnText' ever a empty string?

webapp/src/plugin_constants/icons.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+type SvgIconNames = 'check';
+
+const SVGIcons: Record<SvgIconNames, JSX.Element> = {
+    check: (
+        <>
+            {/* <svg xmlns='http://www.w3.org/2000/svg' width='58' height='58' viewBox='0 0 58 58' fill='none'> */}
```
raghavaggarwal2308, line 8: why is this code commented?

raghavaggarwal2308 (changes requested):

webapp/src/components/resultPanel/index.tsx
ayusht2810: it is an optional prop, so it can be undefined

raghavaggarwal2308: Can be undefined but not empty right?

ayusht2810: updated

raghavaggarwal2308 (approved):

plugin.json
```diff
@@ -43,7 +43,7 @@
                 "key": "EncryptionSecret",
                 "display_name": "Encryption Secret:",
                 "type": "generated",
-                "help_text": "The secret key is used to encrypt and decrypt the OAuth token.\nRegenerating the secret key will cause the loss of data."
+                "help_text": "The secret key is used to encrypt and decrypt the OAuth token.\nRegenerating the secret will require all users to re-connect their accounts to Azure DevOps."
```
avas27JTG, line 46:
```suggestion
                "help_text": "The secret key used to encrypt and decrypt the OAuth token.\nRegenerating the secret will require all users to re-connect their accounts to Azure DevOps."
```

webapp/src/components/resultPanel/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+    >
+        <>
+            {iconClass ? (
+                <i className={`fa result-panel-icon ${iconClass || 'fa-check-circle-o'}`}/>
+            ) : (
+                <SVGWrapper
+                    className='result-panel-icon'
+                    width={58}
+                    height={58}
+                    viewBox='0 0 58 58'
```
avas27JTG, line 37: not reviewing this as this will be reviewed in UI lib. and doing here will lead to do same thing twice

webapp/src/containers/SubscribeModal/index.tsx
```diff
@@ -102,6 +105,12 @@ const SubscribeModal = () => {
         resetModalState();
     };
 
+    // Opens subscription modal
+    const handleSubscriptionModal = () => {
+        resetModalState();
+        dispatch(toggleShowSubscribeModal({isVisible: true, commandArgs: []}));
```
avas27JTG, line 111:
```suggestion
        dispatch(toggleShowSubscribeModal({isVisible: true, commandArgs: []}));
        resetModalState();
```

always do `        dispatch(toggleShowSubscribeModal({isVisible: true, commandArgs: []}));` as first thing to prevent seeing other reset action happens while modal is about to close

webapp/src/containers/SubscribeModal/index.tsx
```diff
@@ -226,6 +240,16 @@ const SubscribeModal = () => {
                         />
                     ))
                 }
+                {
+                    isAnyProjectLinked && showResultPanel &&
```
avas27JTG, line 244: this change was not working fine, I've fixed here https://github.com/Brightscout/mattermost-plugin-azure-devops/pull/63/files#diff-020aea5bc738fb71db78ff41f22a03418982aa42e1fa9d686e5cd5f6e3058d5e

you can see for the betterment

plugin.json
ayusht2810: I guess this already looks fine.

avas27JTG: when we are writing `is` here the meaning is changed 

as by reading `The secret key is used to encrypt and decrypt the OAuth token` it seems like we are telling user what is a `secret key`

but our goal is not that here we are not explaining or describing the secret key itself but we are just saying that this the secret key which is used for encrypt/decrypt

so reading `The secret key used to encrypt and decrypt the OAuth token` just tells what we need

ayusht2810: okay

avas27JTG (approved):

PR #59, opened 2022-08-26
[MI-2081] Add feature to delete and disconnect user on encryption sec…

ayusht2810:
…ret change

### DIFF
--- a/server/plugin/hooks.go
+++ b/server/plugin/hooks.go
@@ -35,7 +35,7 @@ func (p *Plugin) OnConfigurationChange() error {
 	p.setConfiguration(configuration)
 
 	if oldEncryptionSecret != "" && oldEncryptionSecret != p.getConfiguration().EncryptionSecret {
-		if err := p.Store.DeleteUserTokenOnSecretChange(); err != nil {
+		if err := p.Store.DeleteUserTokenOnEncryptionSecretChange(); err != nil {
 			p.API.LogError("Error in deleting Users.", "Error", err.Error())
 			return err
 		}
--- a/server/plugin/hooks.go
+++ b/server/plugin/hooks.go
@@ -35,7 +35,7 @@ func (p *Plugin) OnConfigurationChange() error {
 	p.setConfiguration(configuration)
 
 	if oldEncryptionSecret != "" && oldEncryptionSecret != p.getConfiguration().EncryptionSecret {
-		if err := p.Store.DeleteUserOnSecretChange(); err != nil {
+		if err := p.Store.DeleteUserTokenOnSecretChange(); err != nil {
 			p.API.LogError("Error in deleting Users.", "Error", err.Error())
 			return err
 		}
--- a/server/plugin/hooks.go
+++ b/server/plugin/hooks.go
@@ -26,13 +26,20 @@ func (p *Plugin) OnConfigurationChange() error {
 		return err
 	}
 
-	p.setConfiguration(configuration)
-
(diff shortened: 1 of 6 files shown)


server/store/utils.go
```diff
@@ -145,3 +146,11 @@ func GetKeyHash(key string) string {
 	_, _ = hash.Write([]byte(key))
 	return base64.StdEncoding.EncodeToString(hash.Sum(nil))
 }
+
+func IsValidUserKey(key string) (string, bool) {
+	res := strings.Split(key, "_")
+	if res[0] == constants.UserIDPrefix {
```
Kshitij-Katiyar, line 152: not checking the length of res here

server/store/store.go
```diff
@@ -58,3 +59,35 @@ func (s *Store) Delete(key string) error {
...
+	for {
+		kvList, err := s.api.KVList(page, constants.UsersPerPage)
+		if err != nil {
+			return err
+		}
+		if len(kvList) == 0 {
+			return nil
+		}
+
+		isUserDeleted := false
```
Kshitij-Katiyar, line 74: please explain the need for this flag and why we are updating the pagnumber according to this flag

server/plugin/hooks.go
```diff
@@ -26,13 +26,20 @@ func (p *Plugin) OnConfigurationChange() error {
...
+	oldEncryptionSecret := p.getConfiguration().EncryptionSecret
 	mattermostSiteURL := p.API.GetConfig().ServiceSettings.SiteURL
 	if mattermostSiteURL == nil {
 		return errors.New("plugin requires Mattermost Site URL to be set")
 	}
 	configuration.MattermostSiteURL = *mattermostSiteURL
+	p.setConfiguration(configuration)
+
+	if oldEncryptionSecret != "" && oldEncryptionSecret != p.getConfiguration().EncryptionSecret {
+		if err := p.Store.DeleteUserOnSecretChange(); err != nil {
```
Kshitij-Katiyar, line 38: we are deling whole user or just its token ?

raghavaggarwal2308 (approved):

server/plugin/hooks.go
ayusht2810: we are deleting details related to OAuth, like access token, refresh token

Kshitij-Katiyar: then DeleteUserTokenOnSecretChange is a better name for that function then

Kshitij-Katiyar (approved):

server/plugin/hooks.go
```diff
@@ -26,13 +26,20 @@ func (p *Plugin) OnConfigurationChange() error {
...
+	oldEncryptionSecret := p.getConfiguration().EncryptionSecret
 	mattermostSiteURL := p.API.GetConfig().ServiceSettings.SiteURL
 	if mattermostSiteURL == nil {
 		return errors.New("plugin requires Mattermost Site URL to be set")
 	}
 	configuration.MattermostSiteURL = *mattermostSiteURL
+	p.setConfiguration(configuration)
+
+	if oldEncryptionSecret != "" && oldEncryptionSecret != p.getConfiguration().EncryptionSecret {
+		if err := p.Store.DeleteUserTokenOnSecretChange(); err != nil {
```
avas27JTG, line 38:
```suggestion
		if err := p.Store.DeleteUserTokenOnEncryptionSecretChange(); err != nil {
```

avas27JTG (approved):

PR #61, opened 2022-08-29
[MI-2098]: Refactor phase-1 code base for final release

avas27JTG:
Refactored phase-1 code base for final release


server/plugin/client.go
```diff
@@ -192,6 +192,15 @@ func (c *client) call(basePath, method, path, contentType string, mattermostUser
 		path = baseURL.String() + path
 	}
 
+	// Check refresh token only for APIs other than OAuth
+	if basePath != constants.BaseOauthURL {
+		if IsAccessTokenExpired, refreshToken := c.plugin.IsAccessTokenExpired(mattermostUserID); IsAccessTokenExpired {
```
ayusht2810, line 197: `IsAccessTokenExpired` is this corrected in other PR?

webapp/src/containers/modals/SubscribeModal/index.tsx
```diff
@@ -124,7 +126,7 @@ const SubscribeModal = () => {
         }
     };
 
-    // Observe for the change in redux state after API call and do the required actions
+    // Observe for the change in redux state after API call to create subscription and do the required actions
```
ayusht2810, line 129:
```suggestion
    // Observe for the change in redux state after the API call to create a subscription and do the required actions
```

webapp/src/hooks/useForm.ts
```diff
@@ -52,8 +52,8 @@ const getValidationErrorMessage = (
     }
 };
 
-// Generic hook to handle form fields
-function useForm(initialFormFields: Record<FormFieldNames, ModalFormFieldConfig>) {
+// Genric hook to handle form fields
```
ayusht2810, line 55:
```suggestion
// Generic hook to handle form fields
```

ayusht2810 (approved):

server/plugin/client.go
avas27JTG: yes this fixed in master

server/store/subscriptions.go
```diff
@@ -106,6 +106,7 @@ func (s *Store) GetAllSubscriptions(userID string) ([]*serializers.SubscriptionD
 
 	var subscriptionList []*serializers.SubscriptionDetails
 	for _, subscription := range subscriptions.ByMattermostUserID[userID] {
+		subscription := subscription
```
manojmalik20, line 98: You should also add a comment above this line specifying why you are doing this.

webapp/src/app.tsx
```diff
@@ -61,14 +59,19 @@ const App = (): JSX.Element => {
         getSubscribeModalState(usePlugin.state).visibility,
     ]);
 
-    // Fetch the list of linked projects list
+    // Fetch the list of linked project list
```
manojmalik20, line 62:
```suggestion
    // Fetch the list of linked projects
```

webapp/src/components/card/index.tsx
```diff
@@ -18,61 +18,59 @@ type CardProps = {
...
-                )) }
+                        <i className='fa fa-trash'/>
+                    </button>
+                )}
             </div>
         </div>
-    );
-};
+        <div className='card-body'>
+
```
manojmalik20, line 48:
```suggestion
```

webapp/src/components/dropdown/index.tsx
```diff
@@ -16,7 +16,7 @@ type DropdownProps = {
     error?: boolean | string;
 }
 
-const Dropdown = ({value, placeholder, options, onChange, customOption, loadingOptions, disabled = false, error = '', required}: DropdownProps): JSX.Element => {
+const Dropdown = ({value, placeholder, options, onChange, customOption, loadingOptions, disabled, error, required}: DropdownProps): JSX.Element => {
```
manojmalik20, line 19: Why this change?

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -102,7 +113,7 @@ const ProjectDetails = (projectDetails: ProjectDetails) => {
                 isOpen={showProjectConfirmationModal}
                 onHide={() => setShowProjectConfirmationModal(false)}
                 onConfirm={handleConfirmUnlinkProject}
-                isLoading={usePlugin.getApiState(plugin_constants.pluginApiServiceConfigs.unlinkProject.apiServiceName, projectDetails).isLoading}
+                isLoading={getApiState(plugin_constants.pluginApiServiceConfigs.unlinkProject.apiServiceName, projectDetails).isLoading}
```
manojmalik20, line 116: You should create variables for this above the return statement and use that here.

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -111,9 +122,9 @@ const ProjectDetails = (projectDetails: ProjectDetails) => {
                 isOpen={showSubscriptionConfirmationModal}
                 onHide={() => setShowSubscriptionConfirmationModal(false)}
                 onConfirm={handleConfirmDeleteSubscription}
-                isLoading={usePlugin.getApiState(plugin_constants.pluginApiServiceConfigs.deleteSubscription.apiServiceName, subscriptionToBeDeleted).isLoading}
+                isLoading={getApiState(plugin_constants.pluginApiServiceConfigs.deleteSubscription.apiServiceName, subscriptionToBeDeleted).isLoading}
                 confirmBtnText='Delete'
-                description='Are you sure you want to delete this subscription?'
+                description='Are you sure you want to delete this subscription ?'
```
manojmalik20, line 127: Why this change?

webapp/src/containers/modals/LinkModal/index.tsx
```diff
@@ -0,0 +1,96 @@
+import React, {useEffect} from 'react';
+import {useDispatch} from 'react-redux';
+
```
manojmalik20, line 3: We should remove the empty folders present because of `.gitkeep`. For eg- `containers/action_buttons/.gitkeep`.

webapp/src/containers/modals/LinkModal/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+import {getLinkModalState} from 'selectors';
+
+import usePluginApi from 'hooks/usePluginApi';
+import useForm from 'hooks/useForm';
+import useApiRequestCompletionState from 'hooks/useApiRequestCompletionState';
+
+const LinkModal = () => {
+    const {linkProjectModal} = plugin_constants.form;
+
+    // Hooks
```
manojmalik20, line 19:
```suggestion
```

webapp/src/containers/modals/SubscribeModal/index.tsx
```diff
@@ -24,6 +24,8 @@ import Utils from 'utils';
 import './styles.scss';
 
 const SubscribeModal = () => {
+    const {subscriptionModal} = plugin_constants.form;
```
manojmalik20, line 27:
```suggestion
    const {subscriptionModal as subscriptionModalFields} = plugin_constants.form;
```

webapp/src/containers/modals/TaskModal/index.tsx
```diff
@@ -89,6 +92,7 @@ const TaskModal = () => {
         }
     };
 
+    // Observe for the change in redux state after API call to create task and do the required actions
```
manojmalik20, line 95:
```suggestion
    // Observe the change in redux state after the API call to create task and do the required actions
```

webapp/src/hooks/useForm.ts
```diff
@@ -75,12 +75,12 @@ function useForm(initialFormFields: Record<FormFieldNames, ModalFormFieldConfig>
     const isErrorInFormValidation = (): boolean => {
         let errorFields = {};
         Object.keys(initialFormFields).forEach((field) => {
-            if (initialFormFields[field as FormFieldNames].validations) {
-                Object.keys(initialFormFields[field as FormFieldNames].validations ?? '').forEach((validation) => {
+            if (initialFormFields[field as FormFieldNames]?.validations) {
+                Object.keys(initialFormFields[field as FormFieldNames]?.validations ?? '').forEach((validation) => {
```
manojmalik20, line 79:
```suggestion
                Object.keys(initialFormFields[field as FormFieldNames].validations).forEach((validation) => {
```
Because it is already being checked in the above if condition.

webapp/src/hooks/useForm.ts
```diff
@@ -108,7 +108,7 @@ function useForm(initialFormFields: Record<FormFieldNames, ModalFormFieldConfig>
     };
 
     // Set value for a specific form field
-    const setSpecificFieldValue = (modifiedFormFields: Partial<Record<FormFieldNames, string>>) => {
+    const setSpecificFieldValue = (modifiedFormFields:Partial<Record<FormFieldNames, string>>) => {
```
manojmalik20, line 111: Revert this change. Fix lint.

webapp/src/selectors/index.tsx
```diff
@@ -11,15 +11,15 @@ export const getGlobalModalState = (state: ReduxState): GlobalModalState => {
 };
 
 export const getLinkModalState = (state: ReduxState): LinkProjectModalState => {
```
manojmalik20, line 13: The type of state should be `GlobalState`.

webapp/src/hooks/useForm.ts
avas27JTG: reverted these type of change may came due to sync, also eslint is not giving any error for this

webapp/src/hooks/useForm.ts
avas27JTG: yes but TS is giving error as `Object is possibly 'undefined'`

webapp/src/containers/modals/SubscribeModal/index.tsx
avas27JTG: we cannot use `as` while destructuring
however, we don't need to mention this explicitly as our global type in the `useForm` covers all

webapp/src/containers/modals/LinkModal/index.tsx
avas27JTG: we are writing types for each block as hooks, state variables, etc
why should we remove it?

webapp/src/containers/Rhs/projectDetails/index.tsx
avas27JTG: this may have occurred due to syncing branches

server/plugin/client.go
```diff
@@ -201,6 +201,15 @@ func (c *client) Call(basePath, method, path, contentType string, mattermostUser
...
```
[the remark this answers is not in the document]
avas27JTG: earlier it was small but later Ayush requested to convert to cap as he need to import and mock for test cases

webapp/src/selectors/index.tsx
avas27JTG: `GlobalState` does not contain our plugin state `'plugins-mattermost-plugin-azure-devops'` so we have extended the `GlobalState` in a custom interface as `ReduxState` where we have Global state along with our plugin state

webapp/src/containers/modals/LinkModal/index.tsx
manojmalik20: It just doesn't seem useful. You can keep it if you want.

webapp/src/containers/modals/SubscribeModal/index.tsx
manojmalik20: I was suggesting to change the variable name to make it more clear.
```suggestion
    const {subscriptionModal: subscriptionModalFields} = plugin_constants.form;
```

server/plugin/client.go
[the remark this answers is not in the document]
manojmalik20: I don't understand. This variable's scope will only be inside the `if` block. It will not be accessible outside that block.

manojmalik20 (approved):

server/plugin/client.go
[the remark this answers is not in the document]
avas27JTG: oh my bad I thought you are saying for everywhere even here `c.plugin.IsAccessTokenExpired(mattermostUserID)`

PR #63, opened 2022-09-01
[MI-2109]: Revamped UI and minor fixes

avas27JTG:
Revamped UI and minor fixes


webapp/src/components/card/subscription/styles.scss
```diff
@@ -6,3 +6,11 @@
...
     }
 }
+
+.icon-label {
+    font-size: 16px;
+}
+
+.icon-mm {
+    margin: 0 3px 0 0;
+}
```
ayusht2810, line 16: add eof

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -137,14 +147,16 @@ const ProjectDetails = (projectDetails: ProjectDetails) => {
...
+    }, [getWebsocketEventState(state).isSubscriptionDeleted]);
+
     return (
         <>
-            <BackButton onClick={handleResetProjectDetails}/>
-            <ToggleSwitch
-                active={showAllSubscriptions}
-                onChange={setShowAllSubscriptions}
-                label={plugin_constants.common.ToggleLabel}
-            />
```
ayusht2810, line 147: this part is updated in master

ayusht2810: many changes are not in sync with the master. please check if this is correct

webapp/src/containers/Rhs/projectList/index.tsx
```diff
@@ -78,13 +83,13 @@ const ProjectList = () => {
                     title='Confirm Project Unlink'
                 />
             }
-            {isLoading && <LinearLoader/> }
+            {isLoading && <LinearLoader/>}
             {
                 isSuccess && (
-                    projectsList?.length > 0 ?
+                    sortedProjectList && sortedProjectList.length > 0 ?
```
ayusht2810, line 89: I think we don't need the first condition `sortedProjectList`.

webapp/src/containers/Rhs/projectList/index.tsx
```diff
@@ -78,13 +83,13 @@ const ProjectList = () => {
...
                         <>
                             {
-                                projectsList.map((item: ProjectDetails) => (
+                                sortedProjectList?.map((item: ProjectDetails) => (
```
ayusht2810, line 92:
```suggestion
                                sortedProjectList.map((item: ProjectDetails) => (
```

webapp/src/containers/modals/LinkModal/index.tsx
```diff
@@ -70,7 +70,7 @@ const LinkModal = () => {
             title='Link New Project'
             onHide={resetModalState}
             onConfirm={onConfirm}
-            confirmBtnText='Link new project'
+            confirmBtnText='Link New Project'
```
ayusht2810, line 73: why we have removed `ResultPanel` from this modal?

webapp/src/types/common/form.d.ts
```diff
@@ -2,9 +2,10 @@ type FieldType = 'dropdown' | 'text' | 'timestamp'
 type ValidationTypes = 'isRequired' | 'maxCharLen' | 'minCharLen' | 'regex' | 'regexErrorMessage'
 type ErrorComponents = 'LinkProjectModal' | 'CreateTaskModal' | 'SubscribeModal'
 
-type SubscriptionModalFields = 'organization' | 'project' | 'eventType' | 'channelID' | 'timestamp'
-type CreateTaskModalFields = 'organization' | 'project' | 'type' | 'title' | 'description' | 'areaPath' | 'timestamp';
 type LinkProjectModalFields = 'organization' | 'project' | 'timestamp'
+type CreateTaskModalFields = 'organization' | 'project' | 'type' | 'title' | 'description' | 'areaPath' | 'timestamp'
+type SubscriptionModalFields = 'organization' | 'project' | 'eventType' | 'channelID' | 'timestamp'
+type ErrorComponents = 'SubscribeModal' | 'CreateTaskModal'
```
ayusht2810, line 8: it is present twice

webapp/src/utils/index.ts
```diff
@@ -6,7 +6,7 @@ import Constants from 'plugin_constants';
 import {getOrganizationList, getProjectList} from './filterData';
 import getErrorMessage from './errorHandling';
 
-const getBaseUrls = (): {pluginApiBaseUrl: string; mattermostApiBaseUrl: string} => {
+const getBaseUrls = (): { pluginApiBaseUrl: string; mattermostApiBaseUrl: string } => {
```
ayusht2810, line 9:
```suggestion
const getBaseUrls = (): {pluginApiBaseUrl: string; mattermostApiBaseUrl: string} => {
```

webapp/src/components/card/subscription/index.tsx
```diff
@@ -5,27 +5,33 @@ import IconButton from 'components/buttons/iconButton';
...
+                        >
+                            {plugin_constants.SVGIcons.workEvent}
+                        </SVGWrapper>
+                    }
+                    labelExtraClassName='margin-left-5'
+                    value={plugin_constants.common.boardsEventTypeMap[eventType as EventType]}
                 />
                 <LabelValuePair
-                    label='Channel'
+                    labelIconClassName={`icon ${channelType === 'P' ? 'icon-lock-outline' : 'icon-globe'} icon-label`}
```
sooraj-shukla, line 34: Add `channelType` in constants.

webapp/src/components/card/subscription/styles.scss
sooraj-shukla, line 16: No newline at the end of file

webapp/src/utils/filterData.ts
```diff
@@ -19,4 +19,9 @@ export const getProjectList = (data: ProjectDetails[]) => {
     return projectList;
 };
 
-export const getCurrentChannelSubscriptions = (data: SubscriptionDetails[], channelID: string): SubscriptionDetails[] => data.filter(((subscription) => subscription.channelID === channelID));
```
ayusht2810, line 22: sync with master

webapp/src/components/toggleSwitch/index.tsx
```diff
@@ -15,10 +14,10 @@ const ToggleSwitch = ({
     active,
     onChange,
     label,
-    labelPositioning = ToggleSwitchLabelPositioning.Left,
```
sooraj-shukla, line 18: Why this change?

webapp/src/components/toggleSwitch/index.tsx
```diff
@@ -15,10 +14,10 @@ const ToggleSwitch = ({
...
+    labelPositioning = 'left',
 }: ToggleSwitchProps): JSX.Element => (
-    <div className={`toggle-switch-container d-flex align-items-center ${labelPositioning === ToggleSwitchLabelPositioning.Right && 'flex-row-reverse'}`}>
-        {label && <span className={labelPositioning === ToggleSwitchLabelPositioning.Left ? 'toggle-switch-label--left' : 'toggle-switch-label--right'}>{label}</span>}
+    <div className={`toggle-switch-container d-flex align-items-center ${labelPositioning === 'right' && 'flex-row-reverse justify-content-end'}`}>
+        {label && <span className={labelPositioning === 'left' ? 'toggle-switch-label--left' : 'toggle-switch-label--right'}>{label}</span>}
```
sooraj-shukla, line 20: Shouldn't we be using constants here for adding default positioning?

webapp/src/styles/_components.scss
```diff
@@ -1,5 +1,5 @@
 .height-rhs {
-    height: 100%;
+    height: calc(100% - 150px); // 100% is the entire height and 150px is the space occupied by the button fixed at the bottom
```
sooraj-shukla, line 2: It'll be good to add `150px` to variables.

webapp/src/styles/_utils.scss
```diff
@@ -34,6 +50,19 @@
     display: flex;
 }
 
+.align-item-center {
```
sooraj-shukla, line 53: There is a util class available in mattermost for this.

webapp/src/styles/_utils.scss
```diff
@@ -53,6 +82,10 @@
     }
 }
 
+.justify-content-end {
+    justify-content: flex-end;
```
sooraj-shukla, line 86: Same as above.

webapp/src/styles/_utils.scss
```diff
@@ -34,6 +50,19 @@
     display: flex;
```
sooraj-shukla, line 50: There is a util class available in mattermost for this.

webapp/src/styles/_utils.scss
```diff
@@ -93,3 +121,9 @@
 .overflow-auto {
     overflow: auto;
 }
+
+.text-truncate {
```
sooraj-shukla, line 125: `text-ellipses` class is used in a few places but that util class isn't there in the util styles.

webapp/src/containers/Rhs/projectDetails/index.tsx
avas27JTG: okay will check when syncing with master

webapp/src/utils/index.ts
```diff
@@ -58,14 +58,24 @@ export const getCreateTaskModalCommandArgs = (arr: Array<string>): CreateTaskCom
...
+export const sortProjectList = (project1: ProjectDetails, project2: ProjectDetails) => {
+    if (project1.projectName.toLocaleLowerCase() < project2.projectName.toLocaleLowerCase()) {
+        return -1;
+    }
+    if (project1.projectName.toLocaleLowerCase() > project2.projectName.toLocaleLowerCase()) {
+        return 1;
+    }
+    return 0;
+};
+
```
sooraj-shukla, line 78:
```suggestion
export const sortProjectList = (project1: ProjectDetails, project2: ProjectDetails) => project1.projectName.toLocaleLowerCase().localeCompare(project2.projectName.toLocaleLowerCase());

```

sooraj-shukla (changes requested):

webapp/src/components/toggleSwitch/index.tsx
avas27JTG: we should use type, not enum as using `type` is easier and bounds the values

webapp/src/components/toggleSwitch/index.tsx
avas27JTG: we can use utilize TS type for this, using constant will lead to always importing that constant while passing as prop whereas type doesn't

webapp/src/styles/_utils.scss
avas27JTG: okay but this might be used at many places so let's not depend on MM utils classes as there might be then utils for other things as well which we don't know

server/plugin/command.go
```diff
@@ -61,9 +62,15 @@ func (p *Plugin) getAutoCompleteData() *model.AutocompleteData {
 	create := model.NewAutocompleteData("boards create [title] [description]", "", "Create a new task")
 	azureDevops.AddCommand(create)
 
-	subscribe := model.NewAutocompleteData("subscribe", "", "Add a subscription")
+	subscribe := model.NewAutocompleteData("boards subscribe", "", "Add a subscription")
```
ayusht2810, line 65:
```suggestion
	subscribe := model.NewAutocompleteData("boards subscribe", "", "Add a board subscription")
```

ayusht2810: maybe? to match with unsubscribe command

webapp/src/components/loader/linear/index.tsx
```diff
@@ -2,8 +2,8 @@ import React from 'react';
 
 import './styles.scss';
 
-const LinearLoader = () : JSX.Element => (
-    <div className='linear-loader'>
+const LinearLoader = ({extraClass}: {extraClass?: string}) : JSX.Element => (
+    <div className={`linear-loader ${extraClass}`}>
```
ayusht2810, line 6: will this class contain undefined if we don't send this prop?

webapp/src/containers/modals/LinkModal/index.tsx
```diff
@@ -30,11 +31,19 @@ const LinkModal = () => {
...
     const resetModalState = (isActionDone?: boolean) => {
         dispatch(toggleShowLinkModal({isVisible: false, commandArgs: [], isActionDone}));
         resetFormFields();
+        setShowResultPanel(false);
+    };
+
+    // Opens link project modal
+    const handleOpenLinkProjectModal = () => {
+        resetModalState();
+        dispatch(toggleShowLinkModal({isVisible: true, commandArgs: []}));
```
ayusht2810, line 46: we should dispatch first?

webapp/src/reducers/index.ts
```diff
@@ -17,7 +17,7 @@ const reducers = combineReducers({
     createTaskModalSlice,
     subscriptionModalSlice,
     projectDetailsSlice,
-    userConnectedSlice,
```
ayusht2810, line 20: shouldn't we remove this reducer as well `userConnectedSlice` if not used anywhere?

webapp/src/utils/filterData.ts
ayusht2810: this is left to be synced.

webapp/src/containers/Rhs/projectDetails/index.tsx
ayusht2810: this is left to be synced.

webapp/src/components/toggleSwitch/index.tsx
sooraj-shukla: Here we're providing value using `enum/constant` instead of hard-coded values. Shouldn't we use `enum`/`constants` in such cases?

webapp/src/styles/_utils.scss
sooraj-shukla: We can see all the utils in `mattermost-webapp`. And these are the things that are provided by `react-bootstrap` also, so we can depend on `react-bootstrap` if not on mm.

webapp/src/components/toggleSwitch/index.tsx
sooraj-shukla: I guess while passing the props it'll be okay to use a string literal. But here in the component, we should use constant because this is something that's not expected to change.

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -109,21 +121,20 @@ const ProjectDetails = (projectDetails: ProjectDetails) => {
...
             if (showAllSubscriptions) {
-                setSubscriptionList(subscriptionData);
+                setSubscriptionList(subscriptionListReturnedByApi);
             } else {
-                setSubscriptionList(getCurrentChannelSubscriptions(subscriptionData, currentChannelId));
+                setSubscriptionList(getCurrentChannelSubscriptions(subscriptionListReturnedByApi, currentChannelId));
             }
         }
-    }, [subscriptionData, showAllSubscriptions]);
+    }, [subscriptionListReturnedByApi]);
```
sooraj-shukla, line 132: Please check if this `useEffect` is running on each render. 
(It should be running because `subscriptionListReturnedByApi` is created again and again in each render cycle. In such cases, we need to memoize the value otherwise the dependency array won't have any performance improvement.)

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -163,22 +176,26 @@ const ProjectDetails = (projectDetails: ProjectDetails) => {
                 description='Are you sure you want to delete this subscription?'
                 title='Confirm Delete Subscription'
             />
-            {isLoading && <LinearLoader/>}
-            <div className='d-flex'>
+            {isLoading && <LinearLoader extraClass='top-0'/>}
```
sooraj-shukla, line 179: `top-0` doesn't specify the meaning of what this class is used for. Can we name it so that it is more descriptive?

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -163,22 +176,26 @@ const ProjectDetails = (projectDetails: ProjectDetails) => {
...
+            <ToggleSwitch
+                active={showAllSubscriptions}
+                onChange={handleToggle}
+                label={'Show All Subscriptions'}
```
sooraj-shukla, line 183: Shouldn't we use constant here?

webapp/src/containers/Rhs/projectList/index.tsx
```diff
@@ -63,10 +66,11 @@ const ProjectList = () => {
 
     const {data, isSuccess, isLoading} = getApiState(plugin_constants.pluginApiServiceConfigs.getAllLinkedProjectsList.apiServiceName);
     const projectsList = data as ProjectDetails[];
+    const sortedProjectList = [...projectsList].sort(sortProjectList);
```
sooraj-shukla, line 69: Is `data` always an array? I think, `data` can be `undefined` as well. And in such case, it'll break.

webapp/src/containers/Rhs/projectList/index.tsx
```diff
@@ -78,13 +82,13 @@ const ProjectList = () => {
...
-            {isLoading && <LinearLoader/> }
+            {isLoading && <LinearLoader/>}
             {
                 isSuccess && (
-                    projectsList?.length > 0 ?
+                    sortedProjectList.length > 0 ?
                         <>
                             {
-                                projectsList.map((item: ProjectDetails) => (
+                                sortedProjectList.map((item: ProjectDetails) => (
```
sooraj-shukla, line 91: We can improve the performance by memoizing the `sortedProjectList`. For now, it'll be re-created on each render and because of that, the whole list of projects will be re-rendered on each render no matter whether the projects are changed or not.

webapp/src/containers/Rhs/projectList/index.tsx
```diff
@@ -99,14 +103,14 @@ const ProjectList = () => {
...
                                     className='plugin-btn no-data__btn btn btn-primary project-list-btn'
                                 >
-                                    {'Link new project'}
+                                    {'Link New Project'}
                                 </button>
                             </div>
                         </> :
                         <EmptyState
                             title='No project linked'
                             subTitle={{text: 'You can link a project by clicking the below button.'}}
```
sooraj-shukla, line 112: I think we should use contants for these.

webapp/src/styles/_components.scss
```diff
@@ -1,5 +1,5 @@
 .height-rhs {
-    height: 100%;
+    height: calc(100% - $rhs-bottom-button-height); // 100% is the entire height and 150px is the space occupied by the button fixed at the bottom
```
sooraj-shukla, line 2:
```suggestion
    height: calc(100% - #{$rhs-bottom-button-height}); // 100% is the entire height and 150px is the space occupied by the button fixed at the bottom
```
I think `calc` won't work this way - `calc(100% - $rhs-bottom-button-height)`.

webapp/src/utils/filterData.ts
```diff
@@ -19,4 +19,9 @@ export const getProjectList = (data: ProjectDetails[]) => {
...
+export const getCurrentChannelSubscriptions = (data: SubscriptionDetails[], channelID: string) => (data || []).filter(((subscription) => subscription.channelID === channelID));
```
sooraj-shukla, line 22: `data` is required and `typeof data is array` then why do we need `(data || []).filter...`? Can we not use `data.filter...`

webapp/src/utils/filterData.ts
```diff
@@ -19,4 +19,9 @@ export const getProjectList = (data: ProjectDetails[]) => {
...
+
+export const getCurrentChannelName = (data: ChannelList[], channelID: string) => {
+    const currentChannel = (data || []).filter(((channel) => channel.id === channelID));
+    return currentChannel[0]?.display_name;
+};
```
sooraj-shukla, line 27:
```suggestion
export const getCurrentChannelName = (data: ChannelList[], channelID: string) => data.find((channel) => channel.id === channelID)?.display_name;
```

It was fixed by @ayusht2810, I think it is not in sync with that branch.

sooraj-shukla (changes requested):

webapp/src/components/toggleSwitch/index.tsx
avas27JTG: we can do that but we should take leverage of TS wherever we can use type and literals we can get rid of import/exports

webapp/src/components/toggleSwitch/index.tsx
avas27JTG: same is the case with type if a type is fixed then it's not going to change, we must utilise TS

webapp/src/styles/_utils.scss
avas27JTG: I agree but as I mentioned we have already defined and used our utils at many places in the project so, it would be better to continue with that for now otherwise it will take time finding utils then using them here

webapp/src/components/loader/linear/index.tsx
avas27JTG: yes but that doesn't affect anything in the UI and using `? : ' '` will be more expensive code-wise

webapp/src/utils/filterData.ts
avas27JTG: this is a change of @ayusht2810 can you look into this and try replacing this with the one which you have fixed or is more recent in any branch

webapp/src/utils/filterData.ts
avas27JTG: this is a change of @ayusht2810 can you look into this and try replacing this with the one which you have fixed or is more recent in any branch

webapp/src/containers/Rhs/projectList/index.tsx
avas27JTG: yes, we can refactor texts later as for now we have just used them like this everywhere

webapp/src/containers/Rhs/projectDetails/index.tsx
avas27JTG: I think it specifies because top-0 means o from top similar to margin-0

what you think for this?

webapp/src/containers/Rhs/projectDetails/index.tsx
avas27JTG: this is fine for now in current branch, please update `getCurrentChannelSubscriptions` here

webapp/src/components/toggleSwitch/index.tsx
sooraj-shukla: Yeah, I understand that. But for providing values, we can't use `type`, right? For that, we'll either have to use `enum` or `constant`, right?

webapp/src/components/toggleSwitch/index.tsx
sooraj-shukla: `type` can be `left | right`, but by default, we want `left` here.

webapp/src/styles/_utils.scss
sooraj-shukla: We can take the help of vscode's global-replace - `Ctrl + shift + H` to easily change it.

webapp/src/components/loader/linear/index.tsx
sooraj-shukla: We can initialize `extraClass = ''` in the starting to overcome both these issues.

webapp/src/components/toggleSwitch/index.tsx
avas27JTG: @sooraj-shukla let's sync on this I think there is come confusion

webapp/src/containers/Rhs/projectDetails/index.tsx
sooraj-shukla: okay, seems fine now. I didn't understand why `top-0` was used.

webapp/src/containers/Rhs/projectDetails/index.tsx
avas27JTG: nice catch

webapp/src/containers/Rhs/projectList/index.tsx
```diff
@@ -62,11 +65,12 @@ const ProjectList = () => {
     });
 
     const {data, isSuccess, isLoading} = getApiState(plugin_constants.pluginApiServiceConfigs.getAllLinkedProjectsList.apiServiceName);
-    const projectsList = data as ProjectDetails[];
+    const projectsList = data as ProjectDetails[] ?? [];
+    const sortedProjectList = useMemo(() => [...projectsList].sort(sortProjectList), [projectsList]);
```
sooraj-shukla, line 69: This will still be recreated on each render since the dependency will change again and again. And hence will force the useEffect(whose dependency is `sortedProjectList`) to run on each render.

We can add a to-do comment to memoize usePluginAPI's methods and then memoize these kinds of `object type` values(which we use in `useEffect's dependency`) or we use `primitive` values in the dependency.

sooraj-shukla (approved):

README.md
```diff
@@ -21,7 +21,7 @@ This plugin integrates the services of Azure DevOps in Mattermost. For a stable
 This plugin contains the following features:
 - Right-hand sidebar (RHS) shows the list of linked projects and subscriptions created for each project with separate buttons to connect an account, link or unlink a project and add or delete a subscription.
 
-- Preview work item: A preview of the work item will be created when a work item URL is posted in a channel.
+- Preview work item: A preview of the work item will be created when a work item URL is posted in a channel and the user account is conected.
```
ayusht2810, line 24:
```suggestion
- Preview work item: A preview of the work item will be created when a work item URL is posted in a channel and the user account is connected.
```

docs/developer_docs.md
```diff
@@ -22,7 +22,7 @@ This plugin integrates the services of Azure DevOps in Mattermost. For a stable
 This plugin contains the following features:
 - Right-hand sidebar (RHS) shows the list of linked projects and subscriptions created for each project with separate buttons to connect an account, link or unlink a project and add or delete a subscription.
 
-- Preview work item: A preview of the work item will be created when a work item URL is posted in a channel.
+- Preview work item: A preview of the work item will be created when a work item URL is posted in a channel and the user account is conected.
```
ayusht2810, line 25:
```suggestion
- Preview work item: A preview of the work item will be created when a work item URL is posted in a channel and the user account is connected.
```

ayusht2810 (approved):

webapp/src/plugin_constants/common.ts
```diff
@@ -6,9 +6,12 @@ export const RightSidebarHeader = 'Azure DevOps';
...
-    Right = 'right',
-}
+export const boardsEventTypeMap: Record<EventType, string> = {
+    create: 'Work Item Created',
+    update: 'Work Item Updated',
+    delete: 'Work Item Deleted',
+};
 
-export const ToggleLabel = 'Show all subscriptions';
+export const channelType = {
```
manojmalik20, line 15: This is not needed. There's already a constant for this in the `mattermost-redux` package.

webapp/src/utils/index.ts
```diff
@@ -58,14 +58,16 @@ export const getCreateTaskModalCommandArgs = (arr: Array<string>): CreateTaskCom
     };
 };
 
-export const onPressingEnterKey = (event: React.KeyboardEvent<HTMLSpanElement>, func: () => void) => {
+export const onPressingEnterKey = (event: React.KeyboardEvent<HTMLSpanElement> | React.KeyboardEvent<SVGSVGElement>, func: () => void) => {
```
manojmalik20, line 61:
```suggestion
export const onPressingEnterKey = (event: React.KeyboardEvent<HTMLSpanElement> | React.KeyboardEvent<SVGElement>, func: () => void) => {
```
Is the type for element here contains "SVG" two times?

webapp/src/websocket/index.ts
```diff
@@ -14,3 +14,9 @@ export function handleDisconnect(store: Store<GlobalState, Action<Record<string,
         store.dispatch(toggleIsConnected(false) as Action);
     };
 }
+
+export function handleSubscriptionDeleted(store: Store<GlobalState, Action<Record<string, unknown>>>) {
+    return (_: WebsocketEventParams) => {
```
manojmalik20, line 19: When this websocket event is received here, are you refetching the subscriptions from the API every time? You should refetch only when the RHS is open and it contains the Azure Devops plugin.
See this: https://github.com/Brightscout/mattermost-plugin-servicenow/blob/9dc472a83c62dd840e462fb7743bdf88efd9577c/webapp/src/websocket/index.ts#L56

webapp/src/components/card/base/styles.scss
```diff
@@ -1,7 +1,7 @@
 .wrapper {
     padding: 15px;
-    border-radius: 8px;
+    border-radius: 5px;
     margin-bottom: 15px;
-    border: 1px solid rgba(var(--center-channel-color-rgb), 0.08);
+    border: 1px solid rgba(63, 67, 80, 0.16);
```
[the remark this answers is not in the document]
avas27JTG: we need to match with MM centric border color like border around slack attachment, also this remains same on every theme so didn't use variable

webapp/src/containers/modals/SubscribeModal/index.tsx
```diff
@@ -216,7 +217,7 @@ const SubscribeModal = () => {
         >
             <>
                 {
-                    (getChannelState().isLoading || getOrganizationAndProjectState().isLoading) && <CircularLoader/>
+                    (isChannelListLoading || isOrganizationAndProjectListLoading) && <CircularLoader/>
                 }
```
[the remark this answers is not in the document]
avas27JTG: this line is modified in later PR

webapp/src/utils/index.ts
avas27JTG: yes, I also noticed this but this is expected
Screenshot from 2022-09-08 15-54-04

webapp/src/websocket/index.ts
avas27JTG: even when websocket is fired we are fetching this only when `ProjectDetails` component is mounted as the API call is triggered from this component only

webapp/src/components/card/base/styles.scss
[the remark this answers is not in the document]
manojmalik20: But MM may change their theme colors in the future. I think we should use variable.

webapp/src/websocket/index.ts
manojmalik20: Suppose a subscription is deleted and a websocket event was fired but at that time, the RHS was not open. So, now when we open the RHS, make sure the API doesn't get called twice. This was happening in ServiceNow. Here, the code structure is different, but still we should test it.

webapp/src/components/card/base/styles.scss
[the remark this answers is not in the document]
avas27JTG: actually, MM itself is not using any variable so, we don't have any variable defined for this shade and we just copied the value from MM

image

docs/developer_docs.md
```diff
@@ -22,7 +22,7 @@ This plugin integrates the services of Azure DevOps in Mattermost. For a stable
 This plugin contains the following features:
 - Right-hand sidebar (RHS) shows the list of linked projects and subscriptions created for each project with separate buttons to connect an account, link or unlink a project and add or delete a subscription.
 
-- Preview work item: A preview of the work item will be created when a work item URL is posted in a channel and the user account is connected.
+- Preview work item: A preview of the work item will be created when a work item URL is posted in a channel and the user account is connected to his Azure DevOps account.
```
manojmalik20, line 25:
```suggestion
- Preview work item: A preview of the work item will be created when a work item URL is posted in a channel and the user is connected to his Azure DevOps account.
```
I think this will be better.

webapp/src/websocket/index.ts
avas27JTG: tested, and fixed this

PR #64, opened 2022-09-02
[MI-2115]: Added slash command to show subscriptions list for current…

avas27JTG:
Added slash command to show subscriptions list for the current channel and unsubscribe any.


server/constants/messages.go
```diff
@@ -14,6 +14,7 @@ const (
 	TaskPreviewMessage   = "State: %s\nAssigned To: %s\nDescription: %s"
 	AlreadyLinkedProject = "This project is already linked."
 	NoProjectLinked      = "No project is linked, please link a project."
+	NoSubscriptionFound  = "No boards subscription exists for this channel"
```
Kshitij-Katiyar, line 17:
```suggestion
	NoSubscriptionFound  = "No board's subscription exists for this channel"
```

server/plugin/command.go
```diff
@@ -64,6 +63,12 @@ func (p *Plugin) getAutoCompleteData() *model.AutocompleteData {
 	subscribe := model.NewAutocompleteData("boards subscribe", "", "Add a subscription")
 	azureDevops.AddCommand(subscribe)
 
+	subscriptions := model.NewAutocompleteData("boards subscriptions", "", "View board's subscriptions in the current channel")
```
Kshitij-Katiyar, line 66: What will be better boards subscriptions or board subscription ?

server/plugin/command.go
```diff
@@ -87,10 +92,69 @@ func azureDevopsAccountConnectionCheck(p *Plugin, c *plugin.Context, commandArgs
...
 		return p.sendEphemeralPostForCommand(commandArgs, p.getConnectAccountFirstMessage())
 	}
-
+	if len(args) > 0 {
+		switch {
+		case args[0] == "subscriptions":
+			return azureDevopsSubscribeCommand(p, c, commandArgs, args...)
+		case args[0] == "unsubscribe":
+			return azureDevopsUnsubscribeCommand(p, c, commandArgs, args...)
+		}
```
Kshitij-Katiyar, line 101: do we need a default case here if args[0] is neither subscriptions nor subscriptions

server/plugin/command.go
```diff
@@ -87,10 +92,69 @@ func azureDevopsAccountConnectionCheck(p *Plugin, c *plugin.Context, commandArgs
...
+		p.API.LogError(constants.FetchSubscriptionListError, "Error", err.Error())
+		return p.sendEphemeralPostForCommand(commandArgs, constants.GenericErrorMessage)
+	}
+
+	isSubscriptionFound := false
+	for _, subscription := range subscriptionList {
+		if subscription.SubscriptionID == args[1] {
+			isSubscriptionFound = true
+			p.sendEphemeralPostForCommand(commandArgs, fmt.Sprintf("Boards subscription with ID: \"%s\" is being deleted", args[1]))
+			_, err := p.Client.DeleteSubscription(subscription.OrganizationName, subscription.SubscriptionID, commandArgs.UserId)
```
Kshitij-Katiyar, line 122:
```suggestion
			if _, err := p.Client.DeleteSubscription(subscription.OrganizationName, subscription.SubscriptionID, commandArgs.UserId); err!=nil {
```

server/plugin/command.go
```diff
@@ -87,10 +92,69 @@ func azureDevopsAccountConnectionCheck(p *Plugin, c *plugin.Context, commandArgs
...
+	if err != nil {
+		p.API.LogError(constants.FetchSubscriptionListError, "Error", err.Error())
+		return p.sendEphemeralPostForCommand(commandArgs, constants.GenericErrorMessage)
+	}
+
+	isSubscriptionFound := false
+	for _, subscription := range subscriptionList {
+		if subscription.SubscriptionID == args[1] {
+			isSubscriptionFound = true
+			p.sendEphemeralPostForCommand(commandArgs, fmt.Sprintf("Boards subscription with ID: \"%s\" is being deleted", args[1]))
```
Kshitij-Katiyar, line 121: check its response or this will give lint error

server/plugin/utils.go
```diff
@@ -206,3 +206,41 @@ func (p *Plugin) IsAnyProjectLinked(mattermostUserID string) (bool, error) {
...
+	if len(subscriptionsList) <= 0 {
+		sb.WriteString(constants.NoSubscriptionFound)
+		return sb.String()
+	}
+
+	sb.WriteString("###### Board subscription(s) for this channel\n")
+	sb.WriteString("| Subscription ID | Organization | Project | Event Type |\n")
+	sb.WriteString("| :-------------- | :----------- | :------ | :--------- |\n")
+
+	noSubscriptionFound := true
```
Kshitij-Katiyar, line 222: why can we have a variable named subscriptionFound and do this exactly opposite of this

webapp/src/components/form/index.tsx
```diff
@@ -35,7 +35,7 @@ const Form = ({fieldConfig: {label, type, validations}, value, optionsList, onCh
             <Input
                 type='text'
                 placeholder={label}
-                value={value}
+                value={value ?? ''}
```
Kshitij-Katiyar, line 38: rather than ?? i think we should use || 
```suggestion
                value={value || ''}
```

webapp/src/websocket/index.ts
```diff
@@ -14,3 +14,10 @@ export function handleDisconnect(store: Store<GlobalState, Action<Record<string,
         store.dispatch(toggleIsConnected(false) as Action);
     };
 }
+
+// TODO: types are already added in a PR so, update that here when synced with that
```
Kshitij-Katiyar, line 18:
```
// TODO: Types are already added in a PR so, update that here when synced with that
```

server/plugin/utils.go
```diff
@@ -206,3 +206,41 @@ func (p *Plugin) IsAnyProjectLinked(mattermostUserID string) (bool, error) {
 func (p *Plugin) getConnectAccountFirstMessage() string {
 	return fmt.Sprintf(constants.ConnectAccountFirst, fmt.Sprintf(constants.ConnectAccount, p.GetPluginURLPath(), constants.PathOAuthConnect))
 }
+
+func (p *Plugin) ParseSubscriptionsToCommandResponse(subscriptionsList []serializers.SubscriptionDetails, channelID string) string {
+	var sb strings.Builder
+
+	if len(subscriptionsList) <= 0 {
```
ayusht2810, line 213:
```suggestion
	if len(subscriptionsList) == 0 {
```

ayusht2810 (approved):

server/constants/messages.go
avas27JTG: I added this before but on Azure portal they are always writing `boards` also we can verify all these text messages from Mike later

server/plugin/command.go
avas27JTG: same as above

server/plugin/command.go
avas27JTG: no, that is laready handled and we are showing `Invalid command parameters. Please use /azuredevops help for more information.`

server/plugin/utils.go
avas27JTG: that will need to add extra check as `!subscriptionFound` which is unnecessary

webapp/src/components/form/index.tsx
avas27JTG: we need to check only for `null/undefined` except that any value is valid here

webapp/src/websocket/index.ts
avas27JTG: this comment meant for removal at the end so, I don't think rephrasing this makes any use

server/constants/messages.go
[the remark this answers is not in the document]
avas27JTG: same as https://github.com/Brightscout/mattermost-plugin-azure-devops/pull/64#discussion_r961492167

server/plugin/utils.go
avas27JTG: there would be no chances for `<` but adding this keeps in safer side as we are considering values only > 0

server/plugin/command.go
```diff
@@ -87,10 +92,69 @@ func azureDevopsAccountConnectionCheck(p *Plugin, c *plugin.Context, commandArgs
...
+			if deleteErr := p.Store.DeleteSubscription(&subscription); deleteErr != nil {
+				p.API.LogError("Error in deleting a subscription", "Error", deleteErr.Error())
+				return p.sendEphemeralPostForCommand(commandArgs, constants.GenericErrorMessage)
+			}
+
+			p.API.PublishWebSocketEvent(
+				constants.WSEventSubscriptionDeleted,
+				nil,
+				&model.WebsocketBroadcast{UserId: commandArgs.UserId},
+			)
```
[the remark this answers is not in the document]
ayusht2810: can't we use a `break` to avoid extra iteration?

ayusht2810 (approved):

server/plugin/command.go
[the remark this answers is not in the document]
avas27JTG: we are already returning on line 138 with the post-message, which breaks the loop

webapp/src/containers/Rhs/index.tsx
```diff
@@ -1,22 +1,24 @@
...
 
     return (
         <div className='overflow-auto height-rhs position-relative padding-16'>
             {
-                !isUserAccountConnected() && <AccountNotLinked/>
+                !isConnected && <AccountNotLinked/>
             }
             {
-                isUserAccountConnected() && (
+                isConnected && (
```
manojmalik20, line 21:
```suggestion
            {!isConnected ? <AccountNotLinked/> : (
```

webapp/src/index.tsx
```diff
@@ -33,6 +33,7 @@ export default class Plugin {
         registry.registerRootComponent(SubscribeModal);
         registry.registerWebSocketEventHandler(`custom_${Constants.common.pluginId}_connect`, handleConnect(store));
         registry.registerWebSocketEventHandler(`custom_${Constants.common.pluginId}_disconnect`, handleDisconnect(store));
+        registry.registerWebSocketEventHandler(`custom_${Constants.common.pluginId}_subscriptionDeleted`, handleSubscriptionDeleted(store));
```
manojmalik20, line 36:
```suggestion
        registry.registerWebSocketEventHandler(`custom_${Constants.common.pluginId}_subscription_deleted`, handleSubscriptionDeleted(store));
```

webapp/src/plugin_constants/common.ts
```diff
@@ -6,3 +6,9 @@ export const RightSidebarHeader = 'Azure DevOps';
 
 export const MMCSRF = 'MMCSRF';
 export const HeaderCSRFToken = 'X-CSRF-Token';
+
+export const boardsEventTypeMap: Record<eventType, string> = {
```
manojmalik20, line 10:
```suggestion
export const boardsEventTypeMap: Record<EventType, string> = {
```

webapp/src/containers/Rhs/index.tsx
avas27JTG: if we do this 
```
{
                isConnected ? (
                    getProjectDetailsState(state).projectID ?
                        <ProjectDetails {...getProjectDetailsState(state)}/> :
                        <ProjectList/>
                ) : <AccountNotLinked/>
            }
```

we would be having a nested ternary operator which is not allowed by MM webapp eslint

manojmalik20 (approved):

PR #73, opened 2022-09-08
[MI-2128]: Added logic to auto select dropdown values based on best scenario and improvements.

avas27JTG:
Added logic to auto select dropdown values based on best scenario and improvements.


ayusht2810 (approved):

webapp/src/containers/modals/SubscribeModal/index.tsx
```diff
@@ -163,23 +163,23 @@ const SubscribeModal = () => {
     useEffect(() => {
         if (isChannelListSuccess && !showResultPanel) {
             setChannelOptions(channelList?.map((channel) => ({
-                label: <span><i className={`icon ${channel.type === 'P' ? 'icon-lock-outline' : 'icon-globe'} dropdown-option-icon`}/>{channel.display_name}</span>,
+                label: <span><i className={`icon ${channel.type === plugin_constants.common.channelType.priivate ? 'icon-lock-outline' : 'icon-globe'} dropdown-option-icon`}/>{channel.display_name}</span>,
```
sooraj-shukla, line 166:
```suggestion
                label: <span><i className={`icon ${channel.type === plugin_constants.common.channelType.private ? 'icon-lock-outline' : 'icon-globe'} dropdown-option-icon`}/>{channel.display_name}</span>,
```

sooraj-shukla (approved):

webapp/src/containers/modals/SubscribeModal/index.tsx
```diff
@@ -41,8 +40,9 @@ const SubscribeModal = () => {
         makeApiRequestWithCompletionStatus,
         state,
     } = usePluginApi();
-    const {visibility} = getSubscribeModalState(state);
-    const {currentTeamId} = useSelector((reduxState: GlobalState) => reduxState.entities.teams);
+    const {visibility, project, organization} = getSubscribeModalState(state);
+    const {currentTeamId} = useSelector((pluginState: GlobalState) => pluginState.entities.teams);
```
manojmalik20, line 44:
```suggestion
    const {currentTeamId} = useSelector((state: GlobalState) => pluginState.entities.teams);
```
I don't think the `pluginState` is the right variable name here.

webapp/src/containers/modals/TaskModal/index.tsx
```diff
@@ -178,9 +178,6 @@ const TaskModal = () => {
             error={showApiErrorMessages(isError, error as ApiErrorResponse)}
         >
             <>
-                {
```
manojmalik20, line 181: Why have you removed the loader from everywhere?

webapp/src/types/common/store.d.ts
```diff
@@ -15,6 +15,7 @@ type GlobalModalState = {
 type GlobalModalActionPayload = {
     isVisible: boolean
     commandArgs: Array<string>
+    args?: Array<string>
```
manojmalik20, line 18: What's the main difference between commandArgs and args here?

avas27JTG: `commandArgs` will come from slash command and args will come from anywhere like dispatching action on button click, etc

webapp/src/containers/modals/SubscribeModal/index.tsx
manojmalik20: And what about the previous variable `reduxState`?

manojmalik20 (approved):

webapp/src/containers/modals/SubscribeModal/index.tsx
avas27JTG: yes can use that

PR #75, opened 2022-09-12
[MI-2135]: Moved filter on the server side to get subscriptions for the requested channel

avas27JTG:
Moved filter on the server side to get subscriptions for the requested channel
Modified web app to support new server changes

### DIFF
--- a/server/constants/constants.go
+++ b/server/constants/constants.go
@@ -45,8 +45,8 @@ const (
 	PathParamTeamID = "team_id"
 
 	// URL query params constants
-	QueryParamProject    = "project"
-	QueryParamChanneldID = "channel_id"
+	QueryParamProject   = "project"
+	QueryParamChannelID = "channel_id"
 
 	// Authorization constants
 	Bearer        = "Bearer"
--- a/server/constants/constants.go
+++ b/server/constants/constants.go
@@ -46,7 +46,7 @@ const (
 
 	// URL query params constants
 	QueryParamProject    = "project"
-	QueryParamChanneldId = "channel_id"
+	QueryParamChanneldID = "channel_id"
 
 	// Authorization constants
 	Bearer        = "Bearer"
--- a/server/constants/constants.go
+++ b/server/constants/constants.go
@@ -45,7 +45,8 @@ const (
 	PathParamTeamID = "team_id"
 
 	// URL query params constants
-	QueryParamProject = "project"
+	QueryParamProject    = "project"
+	QueryParamChanneldId = "channel_id"
 
 	// Authorization constants
 	Bearer        = "Bearer"
--- a/server/plugin/api.go
+++ b/server/plugin/api.go
@@ -281,17 +281,13 @@ func (p *Plugin) handleGetSubscriptions(w http.ResponseWriter, r *http.Request) 
 		return
 	}
 
-	channelID := r.URL.Query().Get(constants.QueryParamChanneldID)
+	channelID := r.URL.Query().Get(constants.QueryParamChannelID)
 	project := r.URL.Query().Get(constants.QueryParamProject)
 	if project != "" {
 		subscriptionByProject := []*serializers.SubscriptionDetails{}
 		for _, subscription := range subscriptionList {
 			if subscription.ProjectName == project {
-				if channelID != "" {
-					if subscription.ChannelID == channelID {
-						subscriptionByProject = append(subscriptionByProject, subscription)
-					}
-				} else {
+				if channelID == "" || subscription.ChannelID == channelID {
 					subscriptionByProject = append(subscriptionByProject, subscription)
 				}
 			}
--- a/server/plugin/api.go
+++ b/server/plugin/api.go
@@ -281,12 +281,19 @@ func (p *Plugin) handleGetSubscriptions(w http.ResponseWriter, r *http.Request) 
 		return
 	}
 
+	channelID := r.URL.Query().Get(constants.QueryParamChanneldId)
 	project := r.URL.Query().Get(constants.QueryParamProject)
 	if project != "" {
 		subscriptionByProject := []*serializers.SubscriptionDetails{}
 		for _, subscription := range subscriptionList {
 			if subscription.ProjectName == project {
-				subscriptionByProject = append(subscriptionByProject, subscription)
+				if channelID != "" {
+					if subscription.ChannelID == channelID {
+						subscriptionByProject = append(subscriptionByProject, subscription)
+					}
+				} else {
+					subscriptionByProject = append(subscriptionByProject, subscription)
+				}
 			}
 		}
 		subscriptionList = subscriptionByProject
--- a/server/plugin/api.go
+++ b/server/plugin/api.go
@@ -281,7 +281,7 @@ func (p *Plugin) handleGetSubscriptions(w http.ResponseWriter, r *http.Request) 
 		return
 	}
 
-	channelID := r.URL.Query().Get(constants.QueryParamChanneldId)
+	channelID := r.URL.Query().Get(constants.QueryParamChanneldID)
 	project := r.URL.Query().Get(constants.QueryParamProject)
 	if project != "" {
 		subscriptionByProject := []*serializers.SubscriptionDetails{}
--- a/webapp/src/containers/Rhs/projectDetails/index.tsx
+++ b/webapp/src/containers/Rhs/projectDetails/index.tsx
@@ -22,8 +22,6 @@ import {getSubscribeModalState, getWebsocketEventState} from 'selectors';
 import usePluginApi from 'hooks/usePluginApi';
 import useApiRequestCompletionState from 'hooks/useApiRequestCompletionState';
 
-import {getCurrentChannelSubscriptions} from 'utils/filterData';
-
 const ProjectDetails = memo((projectDetails: ProjectDetails) => {
     const {projectName, organizationName} = projectDetails;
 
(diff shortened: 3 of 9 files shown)


ayusht2810 (approved):

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -35,14 +33,10 @@ const ProjectDetails = memo((projectDetails: ProjectDetails) => {
...
-    const [subscriptionList, setSubscriptionList] = useState<SubscriptionDetails[]>([]);
     const {currentChannelId} = useSelector((reduxState: GlobalState) => reduxState.entities.channels);
-
-    const project: FetchSubscriptionList = {project: projectName};
-    const {data, isLoading} = getApiState(plugin_constants.pluginApiServiceConfigs.getSubscriptionList.apiServiceName, project);
-    const subscriptionData = data as SubscriptionDetails[];
-
+    const [subscriptionListApiParams, setSubscriptionListApiParams] = useState<FetchSubscriptionList>({
+        project: projectName, channel_id: currentChannelId,
+    });
```
sooraj-shukla, line 39: If we do something like below, we can avoid writing few lines of code -

```suggestion
const [showAllSubscription, setShowAllSubscriptions] = useState(false);
 const subscriptionListApiParams = useMemo<FetchSubscriptionList>(() => ({
        project: projectName,
        channel_id: showAllSubscriptions ? '' : currentChannelId,
    }), [projectName, currentChannelId, showAllSubscriptions]);
```

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -103,32 +97,37 @@ const ProjectDetails = memo((projectDetails: ProjectDetails) => {
...
+                ...subscriptionListApiParams,
+                channel_id: '',
+            });
+        } else {
+            setSubscriptionListApiParams({
+                ...subscriptionListApiParams,
+                channel_id: currentChannelId,
+            });
         }
+    };
```
sooraj-shukla, line 112: Now we can remove this function. And use `setShowAllSubscriptions` directly in the ToggleSwitch change handler.

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -103,32 +97,37 @@ const ProjectDetails = memo((projectDetails: ProjectDetails) => {
...
-        if (subscriptionData) {
-            setShowAllSubscriptions(false);
-            setSubscriptionList(getCurrentChannelSubscriptions(subscriptionData, currentChannelId));
+        if (subscriptionListApiParams.channel_id !== currentChannelId) {
+            setSubscriptionListApiParams({
+                ...subscriptionListApiParams,
+                channel_id: currentChannelId,
+            });
         }
     }, [currentChannelId]);
```
sooraj-shukla, line 132: Now we can modify this like below - 
```
useEffect(() => {
 setShowAllSuggestions(false);
}, [currentChannelId]);
```

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -173,8 +175,8 @@ const ProjectDetails = memo((projectDetails: ProjectDetails) => {
             />
             {isLoading && <LinearLoader extraClass='top-0'/>}
             <ToggleSwitch
-                active={showAllSubscriptions}
-                onChange={setShowAllSubscriptions}
+                active={!subscriptionListApiParams.channel_id}
```
sooraj-shukla, line 178:
```suggestion
                active={showAllSubscriptions}
```

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -173,8 +175,8 @@ const ProjectDetails = memo((projectDetails: ProjectDetails) => {
...
+                onChange={handleToggleShowAllSubscriptions}
```
sooraj-shukla, line 179:
```suggestion
                onChange={setShowAllSubscriptions}
```

sooraj-shukla (changes requested):

webapp/src/containers/Rhs/projectDetails/index.tsx
avas27JTG: this led to calling API to fetch the subscription list twice, but that is handled

sooraj-shukla (approved):

webapp/src/containers/Rhs/projectDetails/index.tsx
sooraj-shukla: I think we could've handled it by removing this `useEffect` and changing the `showAllSubscriptions` state inside `useMemo` (I'm guessing that it could work like below).

```
 const subscriptionListApiParams: FetchSubscriptionList = useMemo<FetchSubscriptionList>(() => {
        if (currentChannelId !== subscriptionListApiParams.channel_id) {
            setShowAllSubscriptions(false);
        }
        return {
            project: projectName,
            channel_id: showAllSubscription && currentChannelId === subscriptionListApiParams.channel_id ? '' : currentChannelId,
        };
    }, [projectName, currentChannelId, showAllSubscription]);
```

server/constants/constants.go
```diff
@@ -45,7 +45,8 @@ const (
 	PathParamTeamID = "team_id"
 
 	// URL query params constants
-	QueryParamProject = "project"
+	QueryParamProject    = "project"
+	QueryParamChanneldID = "channel_id"
```
manojmalik20, line 49:
```suggestion
	QueryParamChannelID = "channel_id"
```

server/plugin/api.go
```diff
@@ -281,12 +281,19 @@ func (p *Plugin) handleGetSubscriptions(w http.ResponseWriter, r *http.Request)
...
 		for _, subscription := range subscriptionList {
 			if subscription.ProjectName == project {
-				subscriptionByProject = append(subscriptionByProject, subscription)
+				if channelID != "" {
+					if subscription.ChannelID == channelID {
+						subscriptionByProject = append(subscriptionByProject, subscription)
+					}
+				} else {
+					subscriptionByProject = append(subscriptionByProject, subscription)
+				}
```
manojmalik20, line 296:
```suggestion
				if channelID == "" || subscription.ChannelID == channelID {
					subscriptionByProject = append(subscriptionByProject, subscription)
				}
```

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -21,27 +21,26 @@ import {getSubscribeModalState, getWebsocketEventState} from 'selectors';
...
 
 const ProjectDetails = memo((projectDetails: ProjectDetails) => {
     const {projectName, organizationName} = projectDetails;
 
-    // Hooks
-    const dispatch = useDispatch();
-    const {makeApiRequestWithCompletionStatus, makeApiRequest, getApiState, state} = usePluginApi();
-
     // State variables
+    const [showAllSubscription, setShowAllSubscriptions] = useState(false);
```
manojmalik20, line 30:
```suggestion
    const [showAllSubscriptions, setShowAllSubscriptions] = useState(false);
```

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -105,31 +104,27 @@ const ProjectDetails = memo((projectDetails: ProjectDetails) => {
...
-        if (subscriptionData) {
-            if (showAllSubscriptions) {
-                setSubscriptionList(subscriptionData);
-            } else {
-                setSubscriptionList(getCurrentChannelSubscriptions(subscriptionData, currentChannelId));
-            }
+        /**
+         * Prevent calling API to fetch subscription list twice on switching channel
+         *
+         * If current channel is changed and "showAllSubscription" was true on the last channel then
```
manojmalik20, line 116:
```suggestion
         * If the current channel is changed and "showAllSubscriptions" was true on the last channel then
```

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -105,31 +104,27 @@ const ProjectDetails = memo((projectDetails: ProjectDetails) => {
...
+        fetchSubscriptionList();
+    }, [subscriptionListApiParams]);
 
-    // Update subscription list on switching channels
     useEffect(() => {
-        if (subscriptionData) {
-            setShowAllSubscriptions(false);
-            setSubscriptionList(getCurrentChannelSubscriptions(subscriptionData, currentChannelId));
-        }
+        setShowAllSubscriptions(false);
```
manojmalik20, line 127: Why are we toggling the "Show all subscriptions" switch if the channel is changed? We're not doing this in ServiceNow.

webapp/src/services/index.ts
```diff
@@ -67,8 +67,9 @@ const azureDevOpsPluginApi = createApi({
         [Constants.pluginApiServiceConfigs.getSubscriptionList.apiServiceName]: builder.query<SubscriptionDetails[], FetchSubscriptionList>({
             query: (params) => ({
                 headers: {[Constants.common.HeaderCSRFToken]: Cookies.get(Constants.common.MMCSRF)},
-                url: `${Constants.pluginApiServiceConfigs.getSubscriptionList.path}${params.project}`,
+                url: `${Constants.pluginApiServiceConfigs.getSubscriptionList.path}`,
```
manojmalik20, line 70:
```suggestion
                url: Constants.pluginApiServiceConfigs.getSubscriptionList.path,
```

webapp/src/containers/Rhs/projectDetails/index.tsx
avas27JTG: this will not work because here
we are accessing `subscriptionListApiParams.channel_id` before initialization

also, if we do handle this, another case would be that when the toggle is on and we try to turn it off then `subscriptionListApiParams` will be modified twice once for `setShowAllSubscriptions(false)` and other for `channel_id: showAllSubscription && currentChannelId === subscriptionListApiParams.channel_id ? '' : currentChannelId` which will again call API twice

manojmalik20 (approved):

webapp/src/containers/Rhs/projectDetails/index.tsx
manojmalik20: I think we should discuss this with Mike also.

PR #76, opened 2022-09-16
[MI-2136]: Implemented pagination on server and webapp

avas27JTG:
Implemented pagination on server and webapp


webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -111,44 +147,39 @@ const ProjectDetails = memo((projectDetails: ProjectDetails) => {
...
     useEffect(() => {
         if (getWebsocketEventState(state).isSubscriptionDeleted) {
-            fetchSubscriptionList();
+            handlePagination(true);
             dispatch(toggleIsSubscriptionDeleted(false));
         }
     }, [getWebsocketEventState(state).isSubscriptionDeleted]);
 
+    useEffect(() => {
+        if (!showAllSubscriptions && subscriptionList?.length) {
```
ayusht2810, line 175: we are already setting `subscriptionList` empty initially. I don't think there is a need for this.
```suggestion
        if (!showAllSubscriptions && subscriptionList.length) {
```

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -21,26 +23,54 @@ import {getSubscribeModalState, getWebsocketEventState} from 'selectors';
...
+        offset: paginationQueryParams.offset,
+        limit: paginationQueryParams.limit,
+    }), [projectName, currentChannelId, showAllSubscriptions, paginationQueryParams]);
 
-    // Hooks
-    const previousState = usePreviousState({currentChannelId});
-    const dispatch = useDispatch();
-    const {makeApiRequestWithCompletionStatus, makeApiRequest, getApiState, state} = usePluginApi();
+    const {data, isLoading} = getApiState(plugin_constants.pluginApiServiceConfigs.getSubscriptionList.apiServiceName, subscriptionListApiParams);
+    const subscriptionListReturnedByApi = data as SubscriptionDetails[] || [];
```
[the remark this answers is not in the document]
avas27JTG: no, because we are not initializing it anywhere else with `[]` so, if the API call is in progress or until we don't have `data` from API it will be `undefined` and doing `...` on `undefined` will lead to error

server/constants/messages.go
```diff
@@ -58,4 +58,5 @@ const (
 	InvalidChannelID                = "Invalid channel ID"
 	DeleteSubscriptionError         = "Error in deleting subscription"
 	GetChannelError                 = "Error in getting channels for team and user"
+	InvalidPaginationQueryParam     = "Invalid value for query param(s) offset or limit"
```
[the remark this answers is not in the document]
avas27JTG: for other validations we have used just `Invalid<name>` putting `Error` at suffix will not be correct here

webapp/src/containers/Rhs/projectDetails/index.tsx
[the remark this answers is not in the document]
ayusht2810: we can keep a check with respect to `isLoading`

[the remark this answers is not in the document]
avas27JTG: its the same thing
and its better to always initialise with empty array whenever data is not defined to tackle many cases at once 

checking for loading particularly will not be ideal as, say if API is broken and it's not loading anymore then data could be undefined or null in that case

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -21,26 +23,57 @@ import {getSubscribeModalState, getWebsocketEventState} from 'selectors';
...
+    const hasMoreSubscriptions = useMemo<boolean>(() => (
+        subscriptionListReturnedByApi?.length !== 0 && subscriptionListReturnedByApi?.length === plugin_constants.common.defaultPageLimit
```
ayusht2810, line 60:
```suggestion
        subscriptionListReturnedByApi.length !== 0 && subscriptionListReturnedByApi.length === plugin_constants.common.defaultPageLimit
```

ayusht2810 (approved):

server/plugin/api.go
```diff
@@ -275,12 +276,20 @@ func (p *Plugin) handleCreateSubscription(w http.ResponseWriter, r *http.Request
 func (p *Plugin) handleGetSubscriptions(w http.ResponseWriter, r *http.Request) {
 	mattermostUserID := r.Header.Get(constants.HeaderMattermostUserID)
 	subscriptionList, err := p.Store.GetAllSubscriptions(mattermostUserID)
+
```
manojmalik20, line 279: No need for adding this empty line here.

server/constants/constants.go
```diff
@@ -47,6 +47,8 @@ const (
 	// URL query params constants
 	QueryParamProject   = "project"
 	QueryParamChannelID = "channel_id"
+	QueryParamOffset    = "offset"
+	QueryParamLimit     = "limit"
```
manojmalik20, line 51: I think we should take "page" and "per_page" as the query params as they are more simple from a user's point of view or even a frontend dev's POV.

server/plugin/api.go
```diff
@@ -275,12 +276,20 @@ func (p *Plugin) handleCreateSubscription(w http.ResponseWriter, r *http.Request
...
 	if err != nil {
 		p.API.LogError(constants.FetchSubscriptionListError, "Error", err.Error())
 		p.handleError(w, r, &serializers.Error{Code: http.StatusInternalServerError, Message: err.Error()})
 		return
 	}
 
+	offset, limit, err := p.GetOffsetAndLimitFromQueryParams(r)
+	if err != nil {
+		p.API.LogError(constants.InvalidPaginationQueryParam, "Error", err.Error())
+		p.handleError(w, r, &serializers.Error{Code: http.StatusBadRequest, Message: constants.InvalidPaginationQueryParam})
```
manojmalik20, line 289: I think you should just log the error in server logs and shouldn't return any error in the API response. If the query params for pagination are wrong, you can just use the default values for those and log the error in server logs.

webapp/src/components/emptyState/index.tsx
```diff
@@ -14,100 +14,107 @@ type EmptyStatePropTypes = {
...
+                                height='36'
+                                viewBox='0 0 36 36'
+                                fill='none'
+                                xmlns='http://www.w3.org/2000/svg'
+                            >
+                                <path
+                                    d='M4.449 12.4965L27.033 8.1045L15.978 0V3.7155L3.3705 8.8485L0 13.3065V23.571L4.449 24.855V12.4965ZM13.416 31.407L27.3705 36L36 28.638V6.618L27.0345 8.1045V27.5565L4.449 24.855L13.416 36V31.407Z'
+                                    fill='#8E8E8E'
+                                />
+                            </svg>
```
manojmalik20, line 43: You should create separate components for icons.

server/constants/constants.go
avas27JTG: Sure, sounds good for FE dev
but on the server, we can use offset and limit for betterment

server/plugin/utils.go
```diff
@@ -245,21 +245,24 @@ func (p *Plugin) ParseSubscriptionsToCommandResponse(subscriptionsList []*serial
...
+	} else {
+		page = val
+	}
 
-		if limit, err = strconv.Atoi(queryParamLimit); err != nil {
-			return offset, limit, err
-		}
+	val, err := strconv.Atoi(query.Get(constants.QueryParamPerPage))
+	switch {
+	case err != nil || val < 0 || val > constants.DefaultPerPageLimit: // We can keep max limit per page and default limit per page same
```
manojmalik20, line 260: Why are we keeping the default and max value same? This introduces a limitation in the API. And if we have to keep these values same, then we should keep the value high like "50".

manojmalik20 (approved):

server/plugin/utils.go
avas27JTG: increased it to `50`

we can keep the default and max limit the same because I think keeping them separate doesn't make sense 
let's say if web app doesn't put this query-param then the server should return a feasible set of data up to a point it thinks would be feasible and if they put a higher limit then also our server should send up to that feasible number only

so in case of missing or invalid query param, we should send the best feasible data set only which can be send as max limit

PR #79, opened 2022-09-28
[MI-2193]: Made changes on server to support filters and refactored all the commands

avas27JTG:
- Added flags on the server in the list subscription slash command to support filters.
- Refactored all the commands with nested sub-commands and auto-suggestions.

### DIFF
--- a/server/constants/constants.go
+++ b/server/constants/constants.go
@@ -18,10 +18,20 @@ const (
 		"* `/azuredevops disconnect` - Disconnect your Mattermost account from your Azure DevOps account.\n" +
 		"* `/azuredevops link [projectURL]` - Link your project to a current channel.\n" +
 		"* `/azuredevops boards create [title] [description]` - Create a new task for your project.\n" +
-		"* `/azuredevops subscribe` - Create subscriptions to track changes in tasks for your linked projects.\n" +
-		"* `/azuredevops boards subscriptions [me or anyone] [current-channel or all-channel]` - View board's subscriptions. \n" +
-		"* `/azuredevops boards unsubscribe [subscription id]` - Unsubscribe a board subscription"
-	InvalidCommand = "Invalid command parameters. Please use `/azuredevops help` for more information."
+		"* `/azuredevops boards subscription add` - Add a new Board's subscription for your linked projects.\n" +
+		"* `/azuredevops boards subscription list [me or anyone] [all-channels]` - View Board's subscriptions.\n" +
+		"* `/azuredevops boards subscription unsubscribe [subscription id]` - Unsubscribe a Board's subscription"
+	InvalidCommand      = "Invalid command parameters. Please use `/azuredevops help` for more information."
+	CommandHelp         = "help"
+	CommandConnect      = "connect"
+	CommandDisconnect   = "disconnect"
+	CommandLink         = "link"
+	CommandBoards       = "boards"
+	CommandCreate       = "create"
+	CommandSubscription = "subscription"
+	CommandAdd          = "add"
+	CommandList         = "list"
+	CommandUnsubscribe  = "unsubscribe"
 
 	// Get task link preview constants
 	HTTPS              = "https:"
@@ -56,7 +66,7 @@ const (
 	// Filters
 	FilterCreatedByMe     = "me"
 	FilterCreatedByAnyone = "anyone"
-	FilterAllChannel      = "all-channel"
+	FilterAllChannels     = "all-channels"
 	FilterCurrentChannel  = "current-channel"
 
 	DefaultPage         = 0
--- a/server/constants/constants.go
+++ b/server/constants/constants.go
@@ -18,7 +18,9 @@ const (
 		"* `/azuredevops disconnect` - Disconnect your Mattermost account from your Azure DevOps account.\n" +
 		"* `/azuredevops link [projectURL]` - Link your project to a current channel.\n" +
 		"* `/azuredevops boards create [title] [description]` - Create a new task for your project.\n" +
-		"* `/azuredevops subscribe` - Create subscriptions to track changes in tasks for your linked projects.\n"
+		"* `/azuredevops subscribe` - Create subscriptions to track changes in tasks for your linked projects.\n" +
+		"* `/azuredevops boards subscriptions [me or anyone] [current-channel or all-channel]` - View board's subscriptions. \n" +
+		"* `/azuredevops boards unsubscribe [subscription id]` - Unsubscribe a board subscription"
 	InvalidCommand = "Invalid command parameters. Please use `/azuredevops help` for more information."
 
 	// Get task link preview constants
@@ -54,6 +56,8 @@ const (
 	// Filters
 	FilterCreatedByMe     = "me"
 	FilterCreatedByAnyone = "anyone"
+	FilterAllChannel      = "all-channel"
+	FilterCurrentChannel  = "current-channel"
 
 	DefaultPage         = 0
 	DefaultPerPageLimit = 50
--- a/server/constants/constants.go
+++ b/server/constants/constants.go
@@ -19,7 +19,7 @@ const (
 		"* `/azuredevops link [projectURL]` - Link your project to a current channel.\n" +
 		"* `/azuredevops boards create [title] [description]` - Create a new task for your project.\n" +
 		"* `/azuredevops boards subscription add` - Add a new Boards subscription for your linked projects.\n" +
-		"* `/azuredevops boards subscription list [me or anyone] [all-channels]` - View Boards subscriptions.\n" +
+		"* `/azuredevops boards subscription list [me or anyone] [all_channels]` - View Boards subscriptions.\n" +
 		"* `/azuredevops boards subscription delete [subscription id]` - Unsubscribe a Boards subscription"
 	InvalidCommand      = "Invalid command parameters. Please use `/azuredevops help` for more information."
 	CommandHelp         = "help"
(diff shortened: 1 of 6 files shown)


raghavaggarwal2308 (approved):

server/plugin/command.go
```diff
@@ -140,21 +176,30 @@ func azureDevopsUnsubscribeCommand(p *Plugin, c *plugin.Context, commandArgs *mo
...
 	}
 
-	return p.sendEphemeralPostForCommand(commandArgs, fmt.Sprintf("Boards subscription with ID: %q does not exist", args[1]))
+	return p.sendEphemeralPostForCommand(commandArgs, fmt.Sprintf("Boards subscription with ID: %q does not exist", subscriptionIDToBeDeleted))
 }
 
 func azureDevopsListSubscriptionsCommand(p *Plugin, c *plugin.Context, commandArgs *model.CommandArgs, args ...string) (*model.CommandResponse, *model.AppError) {
-	subscriptionList, err := p.Store.GetAllSubscriptions(commandArgs.UserId)
+	// If 4th argument is present then it must be "all-channels"
+	if len(args) >= 4 && args[3] != constants.FilterAllChannels {
```
manojmalik20, line 188: Why use the sign `>=` rather than `==`?

server/plugin/command.go
```diff
@@ -140,21 +176,30 @@ func azureDevopsUnsubscribeCommand(p *Plugin, c *plugin.Context, commandArgs *mo
...
+
+	subscriptionList, err := p.Store.GetAllSubscriptions("")
 	if err != nil {
 		p.API.LogError(constants.FetchSubscriptionListError, "Error", err.Error())
 		return p.sendEphemeralPostForCommand(commandArgs, constants.GenericErrorMessage)
 	}
 
-	return p.sendEphemeralPostForCommand(commandArgs, p.ParseSubscriptionsToCommandResponse(subscriptionList, commandArgs.ChannelId))
+	showForChannelID := commandArgs.ChannelId
+	if len(args) >= 4 && args[3] == constants.FilterAllChannels {
```
manojmalik20, line 199: Same here.

server/plugin/utils.go
```diff
@@ -208,32 +208,41 @@ func (p *Plugin) getConnectAccountFirstMessage() string {
...
 
 	if len(subscriptionsList) == 0 {
 		sb.WriteString(constants.NoSubscriptionFound)
 		return sb.String()
 	}
 
-	sb.WriteString("###### Board subscription(s) for this channel\n")
-	sb.WriteString("| Subscription ID | Organization | Project | Event Type |\n")
-	sb.WriteString("| :-------------- | :----------- | :------ | :--------- |\n")
+	sb.WriteString("###### Board's subscription(s)\n")
```
manojmalik20, line 219:
```suggestion
	sb.WriteString("###### Boards subscription(s)\n")
```
I think there's no need of apostrophe anywhere where you've written `Board's`

server/plugin/utils.go
```diff
@@ -208,32 +208,41 @@ func (p *Plugin) getConnectAccountFirstMessage() string {
...
-				displayEventType = "Work Item Deleted"
+		displayEventType := ""
+		switch {
+		case subscription.EventType == "create":
+			displayEventType = "Work Item Created"
+		case subscription.EventType == "update":
+			displayEventType = "Work Item Updated"
+		case subscription.EventType == "delete":
+			displayEventType = "Work Item Deleted"
+		}
```
manojmalik20, line 233:
```suggestion
		switch subscription.EventType {
		case "create":
			displayEventType = "Work Item Created"
		case "update":
			displayEventType = "Work Item Updated"
		case "delete":
			displayEventType = "Work Item Deleted"
		}
```

webapp/src/utils/index.ts
```diff
@@ -44,19 +44,10 @@ export const getProjectLinkModalArgs = (str: string): LinkPayload => {
...
-    }
-
-    return {
-        title: arr[1] ?? '',
-        description: arr[2] ?? '',
-    };
-};
+export const getCreateTaskModalCommandArgs = (arr: Array<string>): CreateTaskCommandArgs => ({
+    title: arr[1] ?? '',
+    description: arr[2] ?? '',
```
manojmalik20, line 49: What if array has a length less than 3?

server/plugin/command.go
avas27JTG: there could be chances that by mistake user adds an extra argument at the end of the command so, in that case, instead of invalidating the command we should execute it up to valid params

server/plugin/command.go
```diff
@@ -45,31 +46,44 @@ func (ch *Handler) Handle(p *Plugin, c *plugin.Context, commandArgs *model.Comma
...
-	unsubscribe := model.NewAutocompleteData("boards unsubscribe [subscription id]", "", "Unsubscribe a board subscription")
-	azureDevops.AddCommand(unsubscribe)
+	subscription := model.NewAutocompleteData(constants.CommandSubscription, "", "Add/list/unsubscribe subscriptions")
+	subscriptionAdd := model.NewAutocompleteData(constants.CommandAdd, "", "Add a new subscription")
+	subscriptionView := model.NewAutocompleteData(constants.CommandList, "", "List subscriptions")
+	subscriptionUnsubscribe := model.NewAutocompleteData(constants.CommandUnsubscribe, "", "Unsubscribe a subscription")
+	subscriptionUnsubscribe.AddTextArgument("ID of the subscription to be delete", "[subscription id]", "")
+	subscriptionCreatedByMe := model.NewAutocompleteData(constants.FilterCreatedByMe, "", "Created By Me")
+	subscriptionShowForAllChannels := model.NewAutocompleteData(constants.FilterAllChannels, "", "Show for all channels or You can leave this argument to show for the current channel only")
+	subscriptionCreatedByMe.AddCommand(subscriptionShowForAllChannels)
```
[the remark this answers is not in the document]
avas27JTG: textArguments are just suggestions and they are good for variable params like id, uuid, etc
but where param values are fixed their sub-commands are more user-friendly as it allow user to select that value by just hitting enter or selecting that command from suggestion

server/constants/constants.go
```diff
@@ -54,6 +66,8 @@ const (
 	// Filters
 	FilterCreatedByMe     = "me"
 	FilterCreatedByAnyone = "anyone"
+	FilterAllChannels     = "all-channels"
+	FilterCurrentChannel  = "current-channel"
```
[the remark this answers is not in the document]
avas27JTG: I think in command we should use `-` for word breaks, any reason why we should use `_`?

server/constants/constants.go
```diff
@@ -18,8 +18,20 @@ const (
 		"* `/azuredevops disconnect` - Disconnect your Mattermost account from your Azure DevOps account.\n" +
 		"* `/azuredevops link [projectURL]` - Link your project to a current channel.\n" +
 		"* `/azuredevops boards create [title] [description]` - Create a new task for your project.\n" +
-		"* `/azuredevops subscribe` - Create subscriptions to track changes in tasks for your linked projects.\n"
-	InvalidCommand = "Invalid command parameters. Please use `/azuredevops help` for more information."
+		"* `/azuredevops boards subscription add` - Add a new Board's subscription for your linked projects.\n" +
+		"* `/azuredevops boards subscription list [me or anyone] [all-channels]` - View Board's subscriptions.\n" +
+		"* `/azuredevops boards subscription unsubscribe [subscription id]` - Unsubscribe a Board's subscription"
```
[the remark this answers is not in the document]
avas27JTG: have changed the command name to `delete` but I think we should keep the description saying `Unsubscribe` as its more precisely depicting the meaning/use of command and also consistent with RHS

webapp/src/utils/index.ts
avas27JTG: in that case `arr[1], arr[2]` will be `undefined` and we will have empty string

server/constants/constants.go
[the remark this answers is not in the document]
manojmalik20: I was suggesting using `_` just because it is considered one word when we use `_` and two words when we use `-`.

webapp/src/utils/index.ts
manojmalik20: Are you sure you won't get error for accessing indices of an array more than the length?

avas27JTG: Yes, in JS it will not throw any error for this

server/plugin/command.go
```diff
@@ -64,8 +64,8 @@ func (p *Plugin) getAutoCompleteData() *model.AutocompleteData {
 	subscription := model.NewAutocompleteData(constants.CommandSubscription, "", "Add/list/unsubscribe subscriptions")
 	subscriptionAdd := model.NewAutocompleteData(constants.CommandAdd, "", "Add a new subscription")
 	subscriptionView := model.NewAutocompleteData(constants.CommandList, "", "List subscriptions")
-	subscriptionUnsubscribe := model.NewAutocompleteData(constants.CommandUnsubscribe, "", "Unsubscribe a subscription")
-	subscriptionUnsubscribe.AddTextArgument("ID of the subscription to be delete", "[subscription id]", "")
+	subscriptionUnsubscribe := model.NewAutocompleteData(constants.CommandDelete, "", "Unsubscribe a subscription")
+	subscriptionUnsubscribe.AddTextArgument("ID of the subscription to be Deleted", "[subscription id]", "")
```
manojmalik20, line 68:
```suggestion
	subscriptionUnsubscribe.AddTextArgument("ID of the subscription to be deleted", "[subscription id]", "")
```

manojmalik20 (approved):

PR #80, opened 2022-09-30
[MI-2189] Updated UI and APIs to support repos subscription

raghavaggarwal2308:
1. Updated the slack attachments for notification posts on repos subscription.
2. Added "Service Type" field in the subscription modal.
3. Added a field on the subscription card to specify service type.


server/plugin/api.go
```diff
@@ -359,9 +359,105 @@ func (p *Plugin) handleSubscriptionNotifications(w http.ResponseWriter, r *http.
...
+			}
+		}
+
+		attachment = &model.SlackAttachment{
+			Pretext: body.Message.Markdown,
+			Title:   fmt.Sprintf("%d: %s", body.Resource.PullRequestID, body.Resource.Title),
+			Fields: []*model.SlackAttachmentField{
+				{
+					Title: "Target Branch",
+					Value: strings.Split(body.Resource.TargetRefName, "/")[2],
```
ayusht2810, line 388: are we always sure we get length of `body.Resource.TargetRefName` greater than or equal to 3?

server/plugin/api.go
```diff
@@ -359,9 +359,105 @@ func (p *Plugin) handleSubscriptionNotifications(w http.ResponseWriter, r *http.
...
+		for i := 0; i < len(body.Resource.PullRequest.Reviewers); i++ {
+			if i == 0 {
+				reviewers = ""
+			}
+
+			if i != len(body.Resource.PullRequest.Reviewers)-1 {
+				reviewers += fmt.Sprintf("%s, ", body.Resource.PullRequest.Reviewers[i].DisplayName)
+			} else {
+				reviewers += body.Resource.PullRequest.Reviewers[i].DisplayName
+			}
```
ayusht2810, line 414: can we have a common function for this as this is similar to the above case?

server/plugin/api.go
```diff
@@ -359,9 +359,105 @@ func (p *Plugin) handleSubscriptionNotifications(w http.ResponseWriter, r *http.
...
+					Title: "Comment",
+					Value: body.Resource.Comment.Content,
+				},
+			},
+			Footer: body.Resource.PullRequest.Repository.Name,
+		}
+	case constants.CodePushed:
+		commits := "None" // When no commits are present
+		for i := 0; i < len(body.Resource.Commits); i++ {
+			if i == 0 {
```
ayusht2810, line 445: instead of having this condition checked inside the loop every time, we can initialize it with `""` and then check outside the loop `if value == ""`. Isn't this a better approach? check this everywhere.

server/plugin/api.go
```diff
@@ -359,9 +359,105 @@ func (p *Plugin) handleSubscriptionNotifications(w http.ResponseWriter, r *http.
...
+			}
+
+			commits += fmt.Sprintf("\n[%s](%s) : **%s**", body.Resource.Commits[i].CommitID, body.Resource.Commits[i].URL, body.Resource.Commits[i].Comment)
+		}
+
+		attachment = &model.SlackAttachment{
+			Pretext:    body.Message.Markdown,
+			Title:      "Commit(s)",
+			Text:       commits,
+			Footer:     fmt.Sprintf("%s | %s", strings.Split(body.Resource.RefUpdates[0].Name, "/")[2], body.Resource.Repository.Name),
```
ayusht2810, line 456: here also for length, are we sure about its length?

server/plugin/api.go
```diff
@@ -359,9 +359,105 @@ func (p *Plugin) handleSubscriptionNotifications(w http.ResponseWriter, r *http.
...
+			FooterIcon: fmt.Sprintf("%s/plugins/%s/static/%s", p.GetSiteURL(), constants.PluginID, "git-branch-outline.svg"),
```
ayusht2810, line 457: shouldn't we create a constant for this `git-branch-outline.svg`?

server/plugin/api.go
raghavaggarwal2308: yes

server/plugin/api.go
raghavaggarwal2308: Yes

ayusht2810 (approved):

server/plugin/api.go
```diff
@@ -359,9 +375,85 @@ func (p *Plugin) handleSubscriptionNotifications(w http.ResponseWriter, r *http.
...
+	case constants.PullRequestCreated, constants.PullRequestUpdated, constants.PullRequestMerged:
+		reviewers := p.getReviewersListString(body.Resource.Reviewers)
+
+		attachment = &model.SlackAttachment{
+			Pretext: body.Message.Markdown,
+			Title:   fmt.Sprintf("%d: %s", body.Resource.PullRequestID, body.Resource.Title),
+			Fields: []*model.SlackAttachmentField{
+				{
+					Title: "Target Branch",
+					Value: strings.Split(body.Resource.TargetRefName, "/")[2],
```
Kshitij-Katiyar, line 393: i think here we should check whether this array is actually of length 2 or not because it can panic here

webapp/src/containers/modals/SubscribeModal/index.tsx
```diff
@@ -108,6 +112,14 @@ const SubscribeModal = () => {
         }
     };
 
+    useEffect(() => {
+        if (formFields.serviceType === 'board') {
```
Kshitij-Katiyar, line 116: constants for board and repos

server/plugin/api.go
```diff
@@ -338,6 +338,22 @@ func (p *Plugin) handleGetSubscriptions(w http.ResponseWriter, r *http.Request)
...
+			reviewers += reviewersList[i].DisplayName
+		}
+	}
+
+	if reviewers == "" {
+		return "None" // When no reviewers are added
+	}
+	return reviewers
+}
+
```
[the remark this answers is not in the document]
raghavaggarwal2308: In this case we need to put an if condition inside the loop to make the string from `"None"` to `""`

webapp/src/containers/modals/SubscribeModal/index.tsx
raghavaggarwal2308: Fixed [here](https://github.com/Brightscout/mattermost-plugin-azure-devops/pull/81)

avas27JTG (approved):

PR #82, opened 2022-10-04
[MI-2203] Used a shorter commit id for notification post

raghavaggarwal2308:


server/plugin/api.go
```diff
@@ -446,7 +446,7 @@ func (p *Plugin) handleSubscriptionNotifications(w http.ResponseWriter, r *http.
 				commits = ""
 			}
 
-			commits += fmt.Sprintf("\n[%s](%s) : **%s**", body.Resource.Commits[i].CommitID, body.Resource.Commits[i].URL, body.Resource.Commits[i].Comment)
+			commits += fmt.Sprintf("\n[%s](%s) : **%s**", body.Resource.Commits[i].CommitID[0:8], body.Resource.Commits[i].URL, body.Resource.Commits[i].Comment)
```
ayusht2810, line 449: are we always sure about its length?

raghavaggarwal2308: Yes

ayusht2810 (approved):

server/plugin/api.go
avas27JTG, line 449: @raghavaggarwal2308 is this just about truncating the commit-id up to this limit or something else, as on slack screenshot or `git log --oneline` I think we get a different ID?

raghavaggarwal2308: Yes, its just about truncation the commit-id. The smaller commit IDs are also unique in the particular repo, so we can use them to uniquely identify a commit
https://support.configura.com/hc/en-us/articles/360052644053-Are-Git-Commit-id-s-Truncated-in-CET-Developer-

avas27JTG: Okay

avas27JTG (approved):

PR #85, opened 2022-10-06
[MI-2225][Server+Webapp]: Created and integrated filter component and made required changes on the server

avas27JTG:
- Webapp: Created and integrated filter component.
- Server: Made required changes on the server.

image

image

image


webapp/src/hooks/useClickOutside.ts
```diff
@@ -0,0 +1,26 @@
+import {useEffect} from 'react';
+
+/**
+ * Hook that detects clicks outside of the passed ref
+ */
+function useOutsideClick(ref: React.RefObject<HTMLInputElement>, handleOnClickOutside: () => void) {
+    useEffect(() => {
+        /**
+         * Check if clicked on outside of element
+         */
```
raghavaggarwal2308, line 10: Is there a reason that we are not using comments as  `// comment-here`

webapp/src/containers/Rhs/projectDetails/index.tsx
```diff
@@ -37,7 +37,10 @@ const ProjectDetails = memo((projectDetails: ProjectDetails) => {
     });
     const [subscriptionList, setSubscriptionList] = useState<SubscriptionDetails[]>([]);
     const [showAllSubscriptions, setShowAllSubscriptions] = useState(false);
-    const [filter, setFilter] = useState(SubscriptionFilterCreatedBy.me);
+    const [filter, setFilter] = useState<SubscriptionFilters>({
+        createdBy: subscriptionFilters.createdBy.me,
+        serviceType: subscriptionFilters.serviceType.boards,
```
raghavaggarwal2308, line 42: Are we showing only `boards` subscription by default?

raghavaggarwal2308 (changes requested):

webapp/src/containers/Rhs/projectDetails/index.tsx
avas27JTG: flow changed now

raghavaggarwal2308 (approved):

tanisha-agarwal (approved):

server/plugin/api.go
```diff
@@ -329,13 +329,47 @@ func (p *Plugin) handleGetSubscriptions(w http.ResponseWriter, r *http.Request)
...
+							}
+						}
+					case constants.FilterRepos:
+						switch eventType {
+						case "", constants.FilterAll:
+							if subscription.EventType == constants.SubscriptionEventPullRequestCreated ||
+								subscription.EventType == constants.SubscriptionEventPullRequestMerged ||
+								subscription.EventType == constants.SubscriptionEventPullRequestUpdated ||
+								subscription.EventType == constants.SubscriptionEventPullRequestCommented ||
+								subscription.EventType == constants.SubscriptionEventCodePushed {
```
manojmalik20, line 364: I think you should create something like this and use that here
```
var ValidSubscriptionEventsForRepos = map[string]bool {
    SubscriptionEventPullRequestCreated: true,
    SubscriptionEventPullRequestMerged: true,
}
```
You should create two maps for boards and repos in the constants package and then use those maps here.

webapp/src/components/buttons/primaryButton/index.tsx
```diff
@@ -5,12 +5,15 @@ type PrimaryButtonProps = {
     iconName?: string
     extraClass?: string
     onClick?: () => void
+    isSecondaryButton?: boolean
```
manojmalik20, line 8: If this button is getting such a prop for `isSecondaryButton`, then the folder in which this file is present should be renamed. The folder is `primaryButton`, it should be changed to something like `genericButton` or something else.

webapp/src/containers/Rhs/projectDetails/header.tsx
```diff
@@ -13,27 +13,44 @@ import {toggleIsLinkedProjectListChanged} from 'reducers/linkModal';
...
+    const getEventTypeOptions = useCallback((serviceType: string) => {
+        switch (serviceType) {
+        case subscriptionFilters.serviceType.boards:
+            return subscriptionFilterEventTypeBoardsOptions();
+        case subscriptionFilters.serviceType.repo:
+            return subscriptionFilterEventTypeReposOptions();
+        default:
+            return [filterLabelValuePairAll];
+        }
+    }, [filter.serviceType]);
```
manojmalik20, line 52: I don't think the function is depending on `filter.serviceType`.

webapp/src/containers/Rhs/projectDetails/header.tsx
```diff
@@ -54,8 +71,23 @@ const Header = ({projectDetails, showAllSubscriptions, handlePagination, setShow
         },
     });
 
+    const isFilterApplied = useCallback(() => showAllSubscriptions || filter.createdBy !== defaultSubscriptionFilters.createdBy || filter.serviceType !== defaultSubscriptionFilters.serviceType || filter.eventType !== defaultSubscriptionFilters.eventType, [filter, showAllSubscriptions]);
+
     const {isLoading: isUnlinkProjectLoading} = getApiState(plugin_constants.pluginApiServiceConfigs.unlinkProject.apiServiceName, projectDetails);
 
+    // Detects and close the filter popover whenever it is opened and user click outside of it
```
manojmalik20, line 78:
```suggestion
    // Detects and closes the filter popover whenever it is opened and the user clicks outside of it
```

webapp/src/components/buttons/primaryButton/index.tsx
```diff
@@ -5,12 +5,15 @@ type PrimaryButtonProps = {
...
+    isDisabled?: boolean
 }
 
-const PrimaryButton = ({text, iconName, extraClass = '', onClick}: PrimaryButtonProps) => (
+const PrimaryButton = ({text, iconName, extraClass = '', onClick, isSecondaryButton, isDisabled = false}: PrimaryButtonProps) => (
```
manojmalik20, line 12: Component name should also be changed.

webapp/src/plugin_constants/common.ts
```diff
@@ -36,7 +44,34 @@ export const repos = 'repos';
...
-    me: 'me',
-    anyone: 'anyone',
+export const subscriptionFilters = {
+    createdBy: {
+        me: 'me',
+        anyone: 'anyone',
+    },
+    serviceType: {
+        boards: 'boards',
+        repo: 'repos',
```
manojmalik20, line 54:
```suggestion
        repos: 'repos',
```

webapp/src/plugin_constants/form.ts
```diff
@@ -202,13 +202,51 @@ export const linkProjectModal: Record<LinkProjectModalFields, ModalFormFieldConf
...
+        label: 'Repos',
+    },
+    {
+        ...filterLabelValuePairAll,
+    },
+];
+
+export const subscriptionFilterEventTypeBoardsOptions = () => {
+    const options: LabelValuePair[] = [];
+    Object.keys(subscriptionFilters.eventType.boards).forEach((event) => options.push({
```
manojmalik20, line 232:
```suggestion
    Object.keys(subscriptionFilters.eventType.boards).forEach((eventType) => options.push({
```

webapp/src/plugin_constants/form.ts
```diff
@@ -202,13 +202,51 @@ export const linkProjectModal: Record<LinkProjectModalFields, ModalFormFieldConf
...
+        value: event,
+        label: subscriptionFilters.eventType.boards[event as EventTypeBoards],
+    }));
+
+    options.push(filterLabelValuePairAll);
+
```
manojmalik20, line 238:
```suggestion
```

webapp/src/plugin_constants/form.ts
```diff
@@ -202,13 +202,51 @@ export const linkProjectModal: Record<LinkProjectModalFields, ModalFormFieldConf
...
+export const subscriptionFilterEventTypeReposOptions = () => {
+    const options: LabelValuePair[] = [];
+    Object.keys(subscriptionFilters.eventType.repos).forEach((event) => options.push({
+        value: event,
+        label: subscriptionFilters.eventType.repos[event as EventTypeRepos],
+    }));
+
+    options.push(filterLabelValuePairAll);
+
```
manojmalik20, line 250: Same as above.

webapp/src/types/common/index.d.ts
```diff
@@ -1,7 +1,9 @@
 /**
  * Keep all common types here which are to be used throughout the project
 */
-type EventType = 'workitem.created' | 'workitem.updated' | 'workitem.deleted' | 'workitem.commented' | 'git.pullrequest.created'| 'git.pullrequest.updated' | 'ms.vss-code.git-pullrequest-comment-event' | 'git.push' | 'git.pullrequest.merged'
+type EventTypeBoards = 'workitem.created' | 'workitem.updated' | 'workitem.deleted' | 'workitem.commented'
+type EventTypeRepos = 'git.pullrequest.created'| 'git.pullrequest.updated' | 'ms.vss-code.git-pullrequest-comment-event' | 'git.push' | 'git.pullrequest.merged'
```
manojmalik20, line 5: You should use enums for these. You can add a TODO comment if you don't want to do it now.

webapp/src/containers/Rhs/projectDetails/header.tsx
avas27JTG: yes it is, whenever the `serviceType` changes options for this dropdown changes

server/plugin/api.go
```diff
@@ -328,6 +328,21 @@ func (p *Plugin) handleGetSubscriptions(w http.ResponseWriter, r *http.Request)
...
+		constants.SubscriptionEventWorkItemCommented: true,
+	}
+
+	var ValidSubscriptionEventsForRepos = map[string]bool{
+		constants.SubscriptionEventPullRequestCreated:   true,
+		constants.SubscriptionEventPullRequestMerged:    true,
+		constants.SubscriptionEventPullRequestUpdated:   true,
+		constants.SubscriptionEventPullRequestCommented: true,
+		constants.SubscriptionEventCodePushed:           true,
+	}
```
manojmalik20, line 344: Move these maps to constants package.

server/plugin/api.go
```diff
@@ -357,11 +369,7 @@ func (p *Plugin) handleGetSubscriptions(w http.ResponseWriter, r *http.Request)
 					case constants.FilterRepos:
 						switch eventType {
 						case "", constants.FilterAll:
-							if subscription.EventType == constants.SubscriptionEventPullRequestCreated ||
-								subscription.EventType == constants.SubscriptionEventPullRequestMerged ||
-								subscription.EventType == constants.SubscriptionEventPullRequestUpdated ||
-								subscription.EventType == constants.SubscriptionEventPullRequestCommented ||
-								subscription.EventType == constants.SubscriptionEventCodePushed {
+							if validEventType := ValidSubscriptionEventsForRepos[subscription.EventType]; validEventType {
```
manojmalik20, line 372: Same.

webapp/src/containers/Rhs/projectDetails/header.tsx
```diff
@@ -75,7 +75,7 @@ const Header = ({projectDetails, showAllSubscriptions, handlePagination, setShow
 
     const {isLoading: isUnlinkProjectLoading} = getApiState(plugin_constants.pluginApiServiceConfigs.unlinkProject.apiServiceName, projectDetails);
 
-    // Detects and close the filter popover whenever it is opened and user click outside of it
+    // Detects and closes the filter popover whenever it is opened and user click outside of it
```
manojmalik20, line 78:
```suggestion
    // Detects and closes the filter popover whenever it is opened and the user clicks outside of it
```

webapp/src/index.tsx
```diff
@@ -7,7 +7,7 @@ import reducer from 'reducers';
 
 import {handleConnect, handleDisconnect, handleSubscriptionDeleted} from 'websocket';
 
-import {ChannelHeaderBtn} from 'components/buttons/action_buttons';
+import {ChannelHeaderButton} from 'components/buttons/channelHeaderButton';
```
manojmalik20, line 10: I see that some folder names are using snake case and some are using camel case. We should make it consistent.

avas27JTG: this may be happened due to multiple devs working on web app over the time and didn't look on the common practise

server/constants/constants.go
```diff
@@ -104,3 +104,18 @@ const (
 	GitBranchIcon = "git-branch-icon.svg"
 	ProjectIcon   = "project-icon.svg"
 )
+
+var ValidSubscriptionEventsForBoards = map[string]bool{
```
manojmalik20, line 108:
```suggestion
var (
    ValidSubscriptionEventsForBoards = map[string]bool{
```
You can create a block for all the vars.

webapp/src/index.tsx
```diff
@@ -21,7 +21,7 @@ import SubscribeModal from 'containers/modals/SubscribeModal';
 import App from './app';
 
 // eslint-disable-next-line import/no-unresolved
-import {PluginRegistry} from './types/mattermost-webapp';
+import {PluginRegistry} from './types/mattermostWebapp';
```
manojmalik20, line 24: I think we should keep this folder as it is `mattermost-webapp` as it comes the plugin starter template.

manojmalik20 (approved):

webapp/src/index.tsx
avas27JTG: yes but I think to maintain consistency we should keep it the way we have kept others, otherwise keeping two patterns in the same repo will introduce conflicts

PR #87, opened 2022-10-10
[MI-2231] Updated UI of subscription card to display service type and added author name in notification post

raghavaggarwal2308:
Screenshot from 2022-10-10 19-57-45

Screenshot from 2022-10-10 19-57-51


server/plugin/api.go
```diff
@@ -423,8 +425,10 @@ func (p *Plugin) handleSubscriptionNotifications(w http.ResponseWriter, r *http.
 		}
 
 		attachment = &model.SlackAttachment{
-			Pretext: body.Message.Markdown,
-			Title:   fmt.Sprintf("%d: %s", body.Resource.PullRequestID, body.Resource.Title),
+			Pretext:    body.Message.Markdown,
+			AuthorName: "Azure Repos",
```
Kshitij-Katiyar, line 429: create a constant for this and use this for every attachment

webapp/src/components/card/subscription/index.tsx
```diff
@@ -18,7 +17,29 @@ type SubscriptionCardProps = {
 
 const SubscriptionCard = ({handleDeleteSubscrption, subscriptionDetails: {channelType, eventType, serviceType, channelName, createdBy}, subscriptionDetails}: SubscriptionCardProps) => (
     <BaseCard>
-        <div className='d-flex'>
+        <div>
+            <div className='d-flex justify-content-between align-items-center mb-1'>
+                <div className='d-flex'>
```
Kshitij-Katiyar, line 22: do we need 2 d-flex class here ?

raghavaggarwal2308: Yes, one is for the internal component

Kshitij-Katiyar (approved):

webapp/src/styles/_components.scss
```diff
@@ -59,3 +59,13 @@
         color: var(--error-text);
     }
 }
+
+// Do not modify this class, this is a direct class from mattermost-webapp
+// This class is used for the dimensions of author_icon in Mattermost slack attachment
```
avas27JTG, line 64:
```suggestion
// This class is used for the dimensions of author_icon in the Mattermost slack attachment
```

avas27JTG (approved):

PR #93, opened 2022-10-20
[MI-2279][webapp]: Created UI and integrated repository filter for Repos

avas27JTG:
Created UI and integrated repository filter for Repos.

### DIFF
--- a/webapp/src/containers/modals/SubscribeModal/filters/repos.tsx
+++ b/webapp/src/containers/modals/SubscribeModal/filters/repos.tsx
@@ -1,0 +1,53 @@
+import React, {useEffect, useMemo} from 'react';
+
+import pluginConstants from 'pluginConstants';
+
+import Dropdown from 'components/dropdown';
+
+import usePluginApi from 'hooks/usePluginApi';
+import useApiRequestCompletionState from 'hooks/useApiRequestCompletionState';
+
+import {formLabelValuePairs} from 'utils';
+
+type ReposFilterProps = {
+    handleSelectRepo: (repo: string) => void
+    selectedRepo: string
+} & ReposSubscriptionFiltersRequest
+
+const ReposFilter = ({handleSelectRepo, project, organization, selectedRepo}: ReposFilterProps) => {
+    const {
+        getApiState,
+        makeApiRequestWithCompletionStatus,
+    } = usePluginApi();
+
+    const reposSubscriptionFiltersRequest = useMemo<ReposSubscriptionFiltersRequest>(() => ({
+        organization,
+        project,
+    }), [organization, project]);
+
+    useEffect(() => {
+        makeApiRequestWithCompletionStatus(
+            pluginConstants.pluginApiServiceConfigs.getRepositories.apiServiceName,
+            reposSubscriptionFiltersRequest,
+        );
+    }, [reposSubscriptionFiltersRequest]);
+
+    const {data, isLoading: isGetRepositoriesLoading, isError: isGetRepositoriesError} = getApiState(
+        pluginConstants.pluginApiServiceConfigs.getRepositories.apiServiceName,
+        reposSubscriptionFiltersRequest as APIRequestPayload,
+    );
+    const repositoriesData = data as ReposSubscriptionFiltersResponse[] || [];
+
+    return (
+        <Dropdown
+            placeholder='Repository'
+            value={selectedRepo}
+            onChange={handleSelectRepo}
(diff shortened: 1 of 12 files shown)


webapp/src/containers/modals/SubscribeModal/filters/repos.tsx
```diff
@@ -0,0 +1,53 @@
...
+            pluginConstants.pluginApiServiceConfigs.getRepositories.apiServiceName,
+            reposSubscriptionFiltersRequest,
+        );
+    }, [reposSubscriptionFiltersRequest]);
+
+    const {data, isLoading: isGetRepositoriesLoading, isError: isGetRepositoriesError} = getApiState(
+        pluginConstants.pluginApiServiceConfigs.getRepositories.apiServiceName,
+        reposSubscriptionFiltersRequest as APIRequestPayload,
+    );
+    const repositoriesData = data as ReposSubscriptionFiltersResponse[] || [];
```
raghavaggarwal2308, line 39: can data be `null` | `undefined`?

webapp/src/containers/modals/SubscribeModal/index.tsx
```diff
@@ -243,17 +253,33 @@ const SubscribeModal = () => {
...
-                                    isDisabled={isLoading}
-                                />
-                            ))
+                            <>
+                                {
+                                    Object.keys(subscriptionModalFields).map((field) => (
+                                        <Form
+                                            key={subscriptionModalFields[field as SubscriptionModalFields].label}
+                                            fieldConfig={subscriptionModalFields[field as SubscriptionModalFields]}
+                                            value={formFields[field as SubscriptionModalFields] ?? ''}
```
raghavaggarwal2308, line 262: can formFields[...] be undefined or null? If not, then I think we can skip `?? ''`

raghavaggarwal2308 (changes requested):

webapp/src/containers/modals/SubscribeModal/index.tsx
avas27JTG: yes, they can be undefined especially when the API call is in progress or there doesn't exist a default value for a field.

webapp/src/containers/modals/SubscribeModal/filters/repos.tsx
avas27JTG: yes it can be, and the reason we didn't use `??` here is to cover all possible falsy values instead of just `null, undefined`

raghavaggarwal2308 (approved):

webapp/src/containers/modals/SubscribeModal/filters/repos.tsx
```diff
@@ -0,0 +1,53 @@
...
+    }), [organization, project]);
+
+    useEffect(() => {
+        makeApiRequestWithCompletionStatus(
+            pluginConstants.pluginApiServiceConfigs.getRepositories.apiServiceName,
+            reposSubscriptionFiltersRequest,
+        );
+    }, [reposSubscriptionFiltersRequest]);
+
+    const {data, isLoading: isGetRepositoriesLoading, isError: isGetRepositoriesError} = getApiState(
```
manojmalik20, line 35:
```suggestion
    const {data, isLoading: getRepositoriesLoading, isError: getRepositoriesError} = getApiState(
```

webapp/src/utils/index.ts
```diff
@@ -59,9 +59,40 @@ export const onPressingEnterKey = (event: React.KeyboardEvent<HTMLSpanElement> |
...
+
+    let newUrl = url;
+    Object.keys(pathParams).forEach((param) => {
+        newUrl = newUrl.replace(`:${param}`, pathParams[param]);
+    });
+    return newUrl;
+};
+
+export const formLabelValuePair = (labelKey: string, valueKey: string, data: Record<string, string>) => {
+    const labelValuePair: LabelValuePair = {
```
manojmalik20, line 75:
```suggestion
export const formLabelValuePair = (labelKey: string, valueKey: string, data: Record<string, string>): LabelValuePair => ({
```

manojmalik20 (approved):

webapp/src/containers/modals/SubscribeModal/filters/repos.tsx
avas27JTG: I think adding `is` here makes it more readable for a boolean, it's like asking if this is loading which could have either yes or no as a concrete answer

manojmalik20 (approved):

PR #95, opened 2022-10-21
[MI-2284][server+webapp]: Made server changes to add filters as unique identifier and Created UI and integrated target-branch filter support.

avas27JTG:
Made server changes to add filters as unique identifier
Created UI
Integrated target-branch filter support


server/serializers/subscriptions.go
```diff
@@ -71,6 +73,8 @@ type SubscriptionDetails struct {
 	ChannelType      string `json:"channelType"`
 	SubscriptionID   string `json:"subscriptionID"`
 	CreatedBy        string `json:"createdBy"`
+	TargetBranch     string `json:"targetBranch"`
+	Repository       string `json:"repository"`
```
ayusht2810, line 77: are these fields optional? if not we should validate them in a function used for validating payload `IsSubscriptionRequestPayloadValid`

ayusht2810 (approved):

server/serializers/subscriptions.go
avas27JTG: yes these are optional

webapp/src/containers/modals/SubscribeModal/filters/repos.tsx
```diff
@@ -25,28 +27,65 @@ const ReposFilter = ({handleSelectRepo, project, organization, selectedRepo}: Re
...
+    useEffect(() => {
+        if (selectedTargetBranch !== filterLabelValuePairAll.value) {
+            makeApiRequestWithCompletionStatus(
+                pluginConstants.pluginApiServiceConfigs.getRepositoryBranches.apiServiceName,
+                reposSubscriptionTargetBranchFiltersRequest,
+            );
+        }
+    }, [reposSubscriptionTargetBranchFiltersRequest.repository]);
+
+    const {data: repositoriesDateFromApi, isLoading: isGetRepositoriesLoading, isError: isGetRepositoriesError} = getApiState(
```
manojmalik20, line 52:
```suggestion
    const {data: repositoriesDataFromApi, isLoading: isGetRepositoriesLoading, isError: isGetRepositoriesError} = getApiState(
```

webapp/src/utils/errorHandling.ts
```diff
@@ -11,10 +11,10 @@ const getErrorMessage = (
 
     switch (component) {
     case 'SubscribeModal': // Create subscription modal
-        if (errorState.status === 400 && errorState.data.error === pluginConstants.messages.error.subscriptionAlreadyExists) {
-            return errorState.data.error;
+        if (errorState.status === 400 && errorState.data.Error === pluginConstants.messages.error.subscriptionAlreadyExists) {
```
manojmalik20, line 14: Why changed the case of `error`?

avas27JTG: on debugging I found that it's `Error` and not `error` in the response from RTK

server/store/utils.go
```diff
@@ -136,8 +136,8 @@ func GetSubscriptionListMapKey() string {
 	return GetKeyHash(constants.SubscriptionPrefix)
 }
 
-func GetSubscriptionKey(mattermostUserID, projectID, channelID, eventType string) string {
-	return fmt.Sprintf("%s_%s_%s_%s", mattermostUserID, projectID, channelID, eventType)
+func GetSubscriptionKey(mattermostUserID, projectID, channelID, eventType, repository, targetBranch string) string {
+	return fmt.Sprintf("%s_%s_%s_%s_%s_%s", mattermostUserID, projectID, channelID, eventType, repository, targetBranch)
```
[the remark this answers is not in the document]
avas27JTG: We needed a unique combination for a subscription which back then was formed by the filed values `project, channel and event-type` for a user and now 2 more are added to uniquely identify a subscription so it got long.

There could be an alternative way if we use some custom-made unique key, can see that later as a part of optimizations.
Cc: @ayusht2810 (implemented this)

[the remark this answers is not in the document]
manojmalik20: Then please add a TODO comment for this or create a backlog ticket for this.

webapp/src/utils/errorHandling.ts
manojmalik20: Is this because of RTK or Azure API? Do we also need to change this in other plugins?

manojmalik20 (approved):

webapp/src/utils/errorHandling.ts
avas27JTG: this is from RTK, sure I'll confirm after testing once again on all errors

server/store/utils.go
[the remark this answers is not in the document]
avas27JTG: added TODO here, however we are recording all such enhancements here https://docs.google.com/spreadsheets/d/1jyNWP0aPY0Lz3arc8fsSnQEvUBlG2IkDFNJ-1iKremA/edit#gid=431237657

PR #97, opened 2022-10-25
[MI-2288][server+webapp]: Differentiated payload for service-type and Created chip UI for displaying filters on subscription cards.

avas27JTG:
- Differentiated payload for service-type
- Added filed for repo name
- Created chip UI for displaying filters on subscription cards

Screenshots:
image

image


webapp/src/components/card/subscription/index.tsx
```diff
@@ -61,10 +62,29 @@ const SubscriptionCard = ({handleDeleteSubscrption, subscriptionDetails: {channe
...
+                                    {pluginConstants.SVGIcons.cardFilter}
+                                </SVGWrapper>
+                            </div>
+                            <div className='card-chip-wrapper'>
+                                {
+                                    repositoryName && <Chip text={`Repository is ${repositoryName}`}/>
+                                }
+                                {
+                                    targetBranch && <Chip text={`Target Branch is ${targetBranch}`}/>
+                                }
```
tanisha-agarwal, line 83:
```suggestion
                                {
                                    <Chip text={ repositoryName  
                                              ? `Repository is ${repositoryName}` 
                                              : `Target Branch is ${targetBranch}`
                                          }
                                    />
                                }
```

tanisha-agarwal (approved):

webapp/src/components/card/subscription/index.tsx
avas27JTG: If both `repositoryName` and `targetBranch` are present at a time then we need to show both of them, so we need two separate `Chip`. If we use ternary then only one of the two will be shown at a time.

avas27JTG: you can see attached screenshot for this UI case

raghavaggarwal2308 (approved):

webapp/src/components/chip/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+import './styles.scss';
+
+type ChipProps = {
+    text: string
+    extraClass?: string
+}
+
+const Chip = ({text, extraClass = ''}: ChipProps) => (
+    <Tooltip tooltipContent={text}>
+        <div className={`chip Badge__box text-truncate ${extraClass}`}>{text}</div>
```
manojmalik20, line 14: Why `Badge__box` class had different casing than other classes?

manojmalik20 (approved):

webapp/src/components/chip/index.tsx
avas27JTG: Yes this is inconsistent naming but it's from Mattermost webapp we have not created this, using directly from MM styling

PR #98, opened 2022-10-27
[MI-2294][server]: Created plugin API to fetch possible values for subscription filters.

avas27JTG:
Created plugin API to fetch possible values for subscription filters


server/plugin/api.go
```diff
@@ -787,6 +788,36 @@ func (p *Plugin) handleGetGitRepositoryBranches(w http.ResponseWriter, r *http.R
...
+
+	body, err := serializers.GetSubscriptionFilterPossibleValuesRequestPayloadFromJSON(r.Body)
+	if err != nil {
+		p.API.LogError("Error in decoding the body for fetching subscription filter possible values", "Error", err.Error())
+		p.handleError(w, r, &serializers.Error{Code: http.StatusBadRequest, Message: err.Error()})
+		return
+	}
+
+	if validationErr := body.IsSubscriptionRequestPayloadValid(); validationErr != nil {
+		p.handleError(w, r, &serializers.Error{Code: http.StatusBadRequest, Message: validationErr.Error()})
```
ayusht2810, line 802: don't we need to log an error here?

server/plugin/client.go
```diff
@@ -236,6 +237,37 @@ func (c *client) GetGitRepositoryBranches(organization, projectName, repository,
...
+	var subscriptionFilters []serializers.SubscriptionFilter
+	for _, filter := range request.Filters {
+		subscriptionFilters = append(subscriptionFilters, serializers.SubscriptionFilter{InputID: filter})
+	}
+
+	subscriptionFiltersRequest := &serializers.GetSubscriptionFilterValuesRequestPayloadFromClient{
+		Subscription: &serializers.CreateSubscriptionBodyPayload{
+			PublisherID:      "tfs",
+			ConsumerID:       "webHooks",
+			ConsumerActionID: "httpRequest",
```
ayusht2810, line 252: i guess there are already constants present for these. please check once

raghavaggarwal2308 (approved):

server/plugin/api.go
```diff
@@ -787,6 +788,36 @@ func (p *Plugin) handleGetGitRepositoryBranches(w http.ResponseWriter, r *http.R
...
+	}
+
+	subscriptionFilterValues, statusCode, err := p.Client.GetSubscriptionFilterPossibleValues(body, mattermostUserID)
+	if err != nil {
+		p.API.LogError(constants.ErrorFetchSubscriptionFilterPossibleValues, "Error", err.Error())
+		p.handleError(w, r, &serializers.Error{Code: statusCode, Message: err.Error()})
+		return
+	}
+
+	filterwiseResponse := make(map[string][]serializers.PossibleValues)
```
manojmalik20, line 813:
```suggestion
	filterwiseResponse := make(map[string][]*serializers.PossibleValues)
```

server/plugin/client.go
```diff
@@ -236,6 +237,37 @@ func (c *client) GetGitRepositoryBranches(organization, projectName, repository,
 	return gitBranchesResponse, statusCode, nil
 }
 
+func (c *client) GetSubscriptionFilterPossibleValues(request *serializers.GetSubscriptionFilterPossibleValuesRequestPayload, mattermostUserID string) (*serializers.SubscriptionFilterPossibleValuesResponseFromClient, int, error) {
+	getSubscriptionFilterValuesURL := fmt.Sprintf(constants.GetSubscriptionFilterPossibleValues, request.Organization)
+
+	var subscriptionFilters []serializers.SubscriptionFilter
```
manojmalik20, line 243:
```suggestion
	var subscriptionFilters []*serializers.SubscriptionFilter
```

server/plugin/client.go
```diff
@@ -236,6 +237,37 @@ func (c *client) GetGitRepositoryBranches(organization, projectName, repository,
...
+			PublisherID:      "tfs",
+			ConsumerID:       "webHooks",
+			ConsumerActionID: "httpRequest",
+			EventType:        request.EventType,
+			PublisherInputs: serializers.PublisherInputsGeneric{
+				ProjectID: request.ProjectID,
+			},
+		},
+		InputValues: subscriptionFilters,
+		Scope:       10,
```
manojmalik20, line 259: What's this scope here? Shouldn't this be a constant variable?

server/serializers/subscriptions.go
```diff
@@ -14,6 +14,10 @@ type UserID struct {
 	UniqueName  string `json:"uniqueName"`
 }
 
+type PublisherInputsGeneric struct {
+	ProjectID string `json:"projectId"`
+}
+
 type PublisherInputsBoards struct {
 	ProjectID string `json:"projectId"`
```
manojmalik20, line 22:
```suggestion
	PublisherInputsGeneric
```
See if this works and if it does, do this for repos too.

server/serializers/subscriptions.go
```diff
@@ -45,7 +49,6 @@ type SubscriptionList struct {
 	Count             int                 `json:"count"`
 	SubscriptionValue []SubscriptionValue `json:"value"`
 }
-
```
manojmalik20, line 48: Why this change? Revert it.

server/serializers/subscriptions.go
```diff
@@ -57,6 +60,37 @@ type CreateSubscriptionRequestPayload struct {
...
+	Filters      []string `json:"filters"`
+}
+
+type SubscriptionFilter struct {
+	InputID string `json:"inputId"`
+}
+
+type GetSubscriptionFilterValuesRequestPayloadFromClient struct {
+	Subscription *CreateSubscriptionBodyPayload `json:"subscription"`
+	InputValues  []SubscriptionFilter           `json:"inputValues"`
```
manojmalik20, line 76:
```suggestion
	InputValues  []*SubscriptionFilter           `json:"inputValues"`
```
Wherever we are using array of struct, we should use pointers.

server/serializers/subscriptions.go
avas27JTG: It didn't work due to a mixture of struct and fields
error message: `mixture of field:value and value elements in struct literal`

server/plugin/client.go
avas27JTG: it is not going to be used anywhere else so, I think there is no need for a constant variable, I have added a comment for its role

ayusht2810 (approved):

manojmalik20 (approved):

PR #102, opened 2022-11-01
[MI-2315] Added slash command to add/list pipeline subscriptions

raghavaggarwal2308:


server/plugin/command.go
```diff
@@ -171,6 +176,31 @@ func azureDevopsReposCommand(p *Plugin, c *plugin.Context, commandArgs *model.Co
 	return executeDefault(p, c, commandArgs, args...)
 }
 
+func azureDevopsPipelinesCommand(p *Plugin, c *plugin.Context, commandArgs *model.CommandArgs, args ...string) (*model.CommandResponse, *model.AppError) {
+	// Check if user's Azure DevOps account is connected
```
ayusht2810, line 180:
```suggestion
	// Check if the user's Azure DevOps account is connected
```

ayusht2810 (approved):

server/plugin/command.go
```diff
@@ -171,6 +176,31 @@ func azureDevopsReposCommand(p *Plugin, c *plugin.Context, commandArgs *model.Co
...
+	if len(args) >= 2 && args[0] == constants.CommandSubscription {
+		switch args[1] {
+		case constants.CommandList:
+			// For "list" command there must be at least 3 arguments
+			if len(args) >= 3 && (args[2] == constants.FilterCreatedByMe || args[2] == constants.FilterCreatedByAnyone) {
+				return azureDevopsListSubscriptionsCommand(p, c, commandArgs, constants.CommandPipelines, args...)
+			}
+		case constants.CommandDelete:
+			return azureDevopsUnsubscribeCommand(p, c, commandArgs, constants.CommandPipelines, args...)
+		case constants.CommandAdd:
```
Kshitij-Katiyar, line 196: we are doing nothing in this case?

Kshitij-Katiyar (approved):

server/plugin/command.go
raghavaggarwal2308: This one is handled in webapp code

server/plugin/utils.go
```diff
@@ -251,6 +251,28 @@ func (p *Plugin) ParseSubscriptionsToCommandResponse(subscriptionsList []*serial
...
+		case constants.SubscriptionEventReleaseDeploymentStarted:
+			displayEventType = "Release Deployment Started"
+		case constants.SubscriptionEventRunStageApprovalCompleted:
+			displayEventType = "Run Stage Approval Completed"
+		case constants.SubscriptionEventRunStageStateChanged:
+			displayEventType = "Run Stage State Changed"
+		case constants.SubscriptionEventRunStageWaitingForApproval:
+			displayEventType = "Run Stage Waiting For Approval"
+		case constants.SubscriptionEventRunStateChanged:
+			displayEventType = "Run State Changed"
```
avas27JTG, line 275: instead of these many switch cases we can create a map with cases as keys and respective values

avas27JTG (approved):

PR #104, opened 2022-11-09
[MI-2339][server]: Added support to show pipeline build link preview.

avas27JTG:
Added support to show pipeline build link preview

Screenshot (Pipeline icon will be fixed on syncing with Raghav's changes)
image


server/constants/constants.go
```diff
@@ -115,14 +118,16 @@ const (
 	WSEventSubscriptionDeleted = "subscription_deleted"
 
 	// Colors
-	ReposIconColor  = "#d74f27"
-	BoardsIconColor = "#53bba1"
+	ReposIconColor    = "#d74f27"
+	BoardsIconColor   = "#53bba1"
+	PipelineIconColor = "#4275E4"
```
raghavaggarwal2308, line 123: These colors are present on line 131 as well. I think they are added in two different PRs. We should sync both.

server/serializers/task.go
```diff
@@ -45,12 +45,17 @@ type TaskFieldValue struct {
...
 }
 
 type Href struct {
 	Href string `json:"href"`
 }
 
+type Web struct {
+	Href string `json:"href"`
+}
+
```
raghavaggarwal2308, line 58: both `Web` and `href` has the same field. Can't we use the same struct at both places?

raghavaggarwal2308 (changes requested):

server/plugin/taskLinkPreview.go
```diff
@@ -106,3 +106,51 @@ func (p *Plugin) PostPullRequestPreview(linkData []string, link, userID, channel
...
+				Value: buildDetails.Status,
+				Short: true,
+			},
+		},
+		Footer:     project,
+		FooterIcon: fmt.Sprintf("%s/plugins/%s/static/%s", p.GetSiteURL(), constants.PluginID, constants.FileNameProjectIcon),
+	}
+	model.ParseSlackAttachment(post, []*model.SlackAttachment{attachment})
+
+	return post, ""
```
ayusht2810, line 155:
```suggestion
	}

	model.ParseSlackAttachment(post, []*model.SlackAttachment{attachment})
	return post, ""
```

server/plugin/taskLinkPreview.go
```diff
@@ -106,3 +106,51 @@ func (p *Plugin) PostPullRequestPreview(linkData []string, link, userID, channel
...
+	buildDetails, _, err := p.Client.GetBuildDetails(organization, project, buildID, userID)
+	if err != nil {
+		return nil, ""
+	}
+
+	post := &model.Post{
+		UserId:    userID,
+		ChannelId: channelID,
+	}
+	attachment := &model.SlackAttachment{
```
ayusht2810, line 123:
```suggestion
	}

	attachment := &model.SlackAttachment{
```

ayusht2810 (approved):

raghavaggarwal2308 (approved):

server/plugin/taskLinkPreview.go
```diff
@@ -106,3 +106,52 @@ func (p *Plugin) PostPullRequestPreview(linkData []string, link, userID, channel
...
 }
+
+func (p *Plugin) PostBuildDetailsPreview(linkData []string, link, userID, channelID string) (*model.Post, string) {
+	organization := linkData[3]
+	project := linkData[4]
+	buildID := strings.Split(linkData[6], "&")[0][16:]
+	buildDetails, _, err := p.Client.GetBuildDetails(organization, project, buildID, userID)
+	if err != nil {
+		return nil, ""
+	}
```
manojmalik20, line 117:
```suggestion
	if err != nil {
	        p.API.LogDebug("Error in getting build details from Azure", "Error", err.Error())
		return nil, ""
	}
```

server/plugin/taskLinkPreview.go
```diff
@@ -106,3 +106,52 @@ func (p *Plugin) PostPullRequestPreview(linkData []string, link, userID, channel
 
 	return post, ""
 }
+
+func (p *Plugin) PostBuildDetailsPreview(linkData []string, link, userID, channelID string) (*model.Post, string) {
+	organization := linkData[3]
+	project := linkData[4]
+	buildID := strings.Split(linkData[6], "&")[0][16:]
```
manojmalik20, line 113: Are we sure that we won't get errors like "Array index out of bounds" or similar errors here?

server/plugin/taskLinkPreview.go
```diff
@@ -106,3 +106,52 @@ func (p *Plugin) PostPullRequestPreview(linkData []string, link, userID, channel
...
+		return nil, ""
+	}
+
+	post := &model.Post{
+		UserId:    userID,
+		ChannelId: channelID,
+	}
+
+	attachment := &model.SlackAttachment{
+		AuthorName: "Azure Pipeline",
```
manojmalik20, line 125:
```suggestion
		AuthorName: "Azure Pipelines",
```

server/plugin/taskLinkPreview.go
```diff
@@ -106,3 +106,52 @@ func (p *Plugin) PostPullRequestPreview(linkData []string, link, userID, channel
...
+		AuthorIcon: fmt.Sprintf("%s/plugins/%s/static/%s", p.GetSiteURL(), constants.PluginID, constants.FileNamePipeline), // TODO: update icon file
```
manojmalik20, line 126: You should put the icon in the `public` folder and use this URL - `<siteurl>/plugins/<pluginID>/public/<filename>`

server/plugin/taskLinkPreview.go
```diff
@@ -106,3 +106,52 @@ func (p *Plugin) PostPullRequestPreview(linkData []string, link, userID, channel
...
+				Short: true,
+			},
+			{
+				Title: "Status",
+				Value: buildDetails.Status,
+				Short: true,
+			},
+		},
+		Footer:     project,
+		FooterIcon: fmt.Sprintf("%s/plugins/%s/static/%s", p.GetSiteURL(), constants.PluginID, constants.FileNameProjectIcon),
```
manojmalik20, line 152: Same here.

manojmalik20 (approved):

server/plugin/taskLinkPreview.go
avas27JTG: yes as we would be reaching here only when we would be having a valid link which satisfies all the conditions, we are checking that in a separate func with this regex `BuildDetailsLinkRegex`

server/plugin/taskLinkPreview.go
avas27JTG: we have kept bot images inside the `asset` folder and are hosting that folder as well so I think keeping them here also does the same thing, pls let me know if we have other reasons as well to use `public`

manojmalik20: Keeping files and images in `assets` folder requires us to host the files explicitly by writing some code which must be written in your `api.go` file. But we can remove that code and make use of the code that MM has already written by keeping files in the public folder.

manojmalik20 (approved):

server/plugin/taskLinkPreview.go
avas27JTG: we have logged this ticket to do this separately once everything gets merged to avoid changing this across different branches/conflicts
https://brightscout.atlassian.net/browse/MI-2462

PR #107, opened 2022-11-16
[MI-2341][webapp+server] Added filters for "Build" and "Release" events

raghavaggarwal2308:
1. Added "Release Pipeline", "Approval Status", "Approval Type", "Stage Name" and "Status" for release events. 
3. Added "Build Pipeline" and "Build Status" for build events. 
4. 3. Updated the API to get filter values for above events.


server/constants/constants.go
```diff
@@ -92,12 +92,15 @@ const (
...
-	FilterBoards          = "boards"
-	FilterRepos           = "repos"
+	FilterCreatedByMe            = "me"
+	FilterCreatedByAnyone        = "anyone"
+	FilterAllChannels            = "all_channels"
+	FilterAll                    = "all"
+	FilterBoards                 = "boards"
+	FilterRepos                  = "repos"
+	EventTypeRelease             = "release"
+	FilterIDReleaseDefinitionID  = "releaseDefinitionId"
```
Kshitij-Katiyar, line 102: Is 2 times ID is needed in these variables ?

webapp/src/containers/modals/SubscribeModal/filters/pipelines.tsx
```diff
@@ -0,0 +1,215 @@
...
+    handleSelectApprovalStatus,
+    selectedApprovalStatus,
+    selectedReleaseStatus,
+    handleSelectReleaseStatus,
+}: PipelinesFilterProps) => {
+    const {buildStatusOptions, releaseApprovalTypeOptions, releaseApprovalStatusOptions, releaseStatusOptions, subscriptionFiltersForPipelines, subscriptionFiltersNameForPipelines} = pluginConstants.form;
+
+    const {
+        getApiState,
+        makeApiRequestWithCompletionStatus,
```
Kshitij-Katiyar, line 57: can this be in 1 line

webapp/src/components/card/subscription/index.tsx
```diff
@@ -27,7 +27,7 @@ const SubscriptionCard = ({handleDeleteSubscrption, subscriptionDetails: {channe
 
     return (
         <BaseCard>
-            <>
+            <div>
```
Kshitij-Katiyar, line 30: why this change ?

webapp/src/containers/modals/SubscribeModal/filters/pipelines.tsx
```diff
@@ -0,0 +1,215 @@
...
+                                loadingOptions={isLoading}
+                                disabled={isLoading}
+                            />
+                        </div>
+                    </>
+                )
+            }
+            {
+                eventType === pluginConstants.common.eventTypePipelineKeys.releaseDeploymentApprovalComplete && (
+                    <>
```
Kshitij-Katiyar, line 179: i think there is no need for these just the below div is enough

webapp/src/pluginConstants/form.ts
```diff
@@ -105,6 +105,78 @@ const serviceTypeOptions: LabelValuePair[] = [
...
+    {
+        value: 'Failed',
+        label: 'Failed',
+    },
+    {
+        value: 'Stopped',
+        label: 'Stopped',
+    },
+    {
+        ...filterLabelValuePairAll,
```
Kshitij-Katiyar, line 126: can we change this filterLabelValuePairAll variable name ?

server/constants/constants.go
raghavaggarwal2308: Yes, the first part is prefix and the second one is the name

webapp/src/containers/modals/SubscribeModal/filters/pipelines.tsx
```diff
@@ -0,0 +1,215 @@
...
+    organization: string
+    projectId: string
+    eventType: string
+    selectedBuildPipeline: string
+    handleSelectBuildPipeline: (value: string, name?: string) => void
+    setIsFiltersError: (value: boolean) => void
+    selectedBuildStatus: string
+    handleSelectBuildStatus: (value: string, name?: string) => void
+    handleSelectReleasePipeline: (value: string, name?: string) => void
+    selectedReleasePipeline: string
```
ayusht2810, line 22:
```suggestion
    handleSelectReleasePipeline: (value: string, name?: string) => void
    selectReleasePipeline: string
```

ayusht2810: shouldn't this be like this or vice-versa? if yes, check everywhere

webapp/src/containers/modals/SubscribeModal/filters/pipelines.tsx
```diff
@@ -0,0 +1,215 @@
...
+        makeApiRequestWithCompletionStatus,
+    } = usePluginApi();
+
+    const getSubscriptionFiltersRequest = useMemo<GetSubscriptionFiltersRequest>(() => ({
+        organization,
+        projectId,
+        filters: subscriptionFiltersForPipelines,
+        eventType,
+        releasePipelineId: selectedReleasePipeline,
+    }), [organization, projectId, eventType, subscriptionFiltersForPipelines, selectedBuildPipeline, selectedReleasePipeline]);
```
ayusht2810, line 66: is this `selectedBuildPipeline` required?

webapp/src/containers/modals/SubscribeModal/filters/pipelines.tsx
```diff
@@ -0,0 +1,215 @@
...
+                getSubscriptionFiltersRequest,
+            );
+        }
+    }, [getSubscriptionFiltersRequest]);
+
+    const {data, isLoading, isError, isSuccess} = getApiState(
+        pluginConstants.pluginApiServiceConfigs.getSubscriptionFilters.apiServiceName,
+        getSubscriptionFiltersRequest as APIRequestPayload,
+    );
+    const filtersData = data as GetSubscriptionFiltersResponse || [];
```
ayusht2810, line 81:
```suggestion
    );
    
    const filtersData = data as GetSubscriptionFiltersResponse || [];
```

webapp/src/containers/modals/SubscribeModal/filters/pipelines.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+    useEffect(() => {
+        if (isError && !isSuccess) {
+            setIsFiltersError(true);
+        } else {
+            setIsFiltersError(false);
+        }
```
ayusht2810, line 88:
```suggestion
            setIsFiltersError(true);
            return;
        } 
        
        setIsFiltersError(false);
        }
```

webapp/src/containers/modals/SubscribeModal/filters/pipelines.tsx
```diff
@@ -0,0 +1,215 @@
...
+    }, [isLoading, isError, isSuccess]);
```
ayusht2810, line 89: is `isLoading` required here?

webapp/src/containers/modals/SubscribeModal/filters/pipelines.tsx
```diff
@@ -0,0 +1,215 @@
...
+                                loadingOptions={isLoading}
+                                disabled={isLoading}
+                            />
+                        </div>
+                    </>
+                )
+            }
+            {
+                eventType.includes('release') && (
+                    <>
```
ayusht2810, line 127:
```suggestion
```

ayusht2810: check everywhere

webapp/src/containers/modals/SubscribeModal/index.tsx
```diff
@@ -282,6 +283,48 @@ const SubscribeModal = () => {
             areaPath: newValue,
         });
 
+    const handleSetBuildPipelineFilter = (newValue: string, pipelineName?: string) =>
```
ayusht2810, line 286: `pipelineName` this is not being used anywhere, is this correct?
same for below as well

webapp/src/pluginConstants/form.ts
raghavaggarwal2308: I think what we have is fine. Do you have any suggestions?

webapp/src/containers/modals/SubscribeModal/filters/pipelines.tsx
raghavaggarwal2308: No, we are storing the selected value in this variable and `handleSelectReleasePipeline` is used when we select a value from the dropdowns

webapp/src/containers/modals/SubscribeModal/filters/pipelines.tsx
raghavaggarwal2308: I think it is used to check for error when loading is completed. Right @avas27JTG ?

server/constants/constants.go
Kshitij-Katiyar: ok then i think the last ID is not needed like for the filter for boards we have filterBoards not filterBoardsFilter

webapp/src/pluginConstants/form.ts
Kshitij-Katiyar: can you put all in between the words like filterLabelPairAllValue or filterLabelPairValues. if that doesn't change the meaning

Kshitij-Katiyar (approved):

webapp/src/containers/modals/SubscribeModal/filters/pipelines.tsx
ayusht2810: on completion of loading, either the `isError` or `isSuccess` will change. So, I don't think we require that

ayusht2810 (approved):

webapp/src/pluginConstants/form.ts
raghavaggarwal2308: Yes, it changes the meaning. The meaning of the name is that it's a label-value pair that contains the value "all"

server/constants/constants.go
raghavaggarwal2308: Done

webapp/src/containers/modals/SubscribeModal/filters/pipelines.tsx
avas27JTG: yes we can remove `isLoading`

avas27JTG (approved):

ayusht2810 (approved):

PR #117, opened 2022-11-28
[MI-2401][webapp]: Fixed MM v5.37 UI bugs and filter API call issues

avas27JTG:
Fixed MM v5.37 UI bugs and filter API call issues

### DIFF
--- a/webapp/src/components/dropdown/index.tsx
+++ b/webapp/src/components/dropdown/index.tsx
@@ -5,7 +5,7 @@ import './styles.scss';
 type DropdownProps = {
     value: string | null;
     placeholder: string;
-    onChange: (newValue: string, targetLabel?: string) => void;
+    onChange: (newValue: string, targetLabel?: string, selectedOption?: Record<string, string>) => void;
     options: LabelValuePair[];
     customOption?: LabelValuePair & {
         onClick: (customOptionValue: string) => void;
@@ -26,7 +26,7 @@ const Dropdown = ({value, placeholder, options, onChange, customOption, loadingO
 
         // Trigger onChange only if there is a change in the dropdown value
         if (newOption.value !== value) {
-            onChange(newOption.value, newOption.label as string);
+            onChange(newOption.value, newOption.label as string, newOption as Record<string, string>);
         }
     };
 
--- a/webapp/src/containers/Rhs/index.tsx
+++ b/webapp/src/containers/Rhs/index.tsx
@@ -1,4 +1,4 @@
-import React, {useEffect} from 'react';
+import React from 'react';
 
 import usePluginApi from 'hooks/usePluginApi';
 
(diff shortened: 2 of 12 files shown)


webapp/src/containers/Rhs/index.tsx
```diff
@@ -13,28 +13,23 @@ import ProjectList from './projectList';
...
-                !isLoading && isConnected && (
-                    getProjectDetailsState(state).projectID ?
-                        <ProjectDetails {...getProjectDetailsState(state)}/> :
-                        <ProjectList/>)
-            }
+            {!isConnected && <AccountNotLinked/>}
+            {isConnected && (
+                getProjectDetailsState(state).projectID ?
+                    <ProjectDetails {...getProjectDetailsState(state)}/> :
+                    <ProjectList/>)}
```
ayusht2810, line 32: can we use the ternary operator here?

ayusht2810 (approved):

webapp/src/containers/modals/SubscribeModal/index.tsx
```diff
@@ -118,15 +119,30 @@ const SubscribeModal = () => {
         }
     };
 
+    const setSelectedDropdownOption = (field: SubscriptionModalFields, newValue: string, selectedOption?: Record<string, string>) => {
+        onChangeFormField(field as SubscriptionModalFields, newValue);
+
+        if (field === 'project' && selectedOption) {
```
raghavaggarwal2308, line 125: Do we need to create a constant for this?

raghavaggarwal2308 (approved):

webapp/src/containers/Rhs/index.tsx
avas27JTG: no as nested ternary not allowed by lint here

webapp/src/containers/modals/SubscribeModal/index.tsx
avas27JTG: no we are using string literal as per the type, however we have added a TODO to create enums in future

webapp/src/containers/modals/SubscribeModal/index.tsx
```diff
@@ -118,15 +119,30 @@ const SubscribeModal = () => {
...
+        }
+    }, [showResultPanel]);
+
     useEffect(() => {
         if (formFields.serviceType === pluginConstants.common.boards) {
-            setSubscriptionModalFields({...subscriptionModalFields, eventType: {...subscriptionModalFields.eventType, optionsList: boardEventTypeOptions}});
+            setSubscriptionModalFields({...subscriptionModalFields, eventType: {...subscriptionModalFields.eventType, optionsList: boardEventTypeOptions, isFieldDisabled: !formFields.project}, serviceType: {...subscriptionModalFields.serviceType, isFieldDisabled: !formFields.project}});
         } else if (formFields.serviceType === pluginConstants.common.repos) {
-            setSubscriptionModalFields({...subscriptionModalFields, eventType: {...subscriptionModalFields.eventType, optionsList: repoEventTypeOptions}});
+            setSubscriptionModalFields({...subscriptionModalFields, eventType: {...subscriptionModalFields.eventType, optionsList: repoEventTypeOptions, isFieldDisabled: !formFields.project}, serviceType: {...subscriptionModalFields.serviceType, isFieldDisabled: !formFields.project}});
```
manojmalik20, line 141:
```
if (formFields.serviceType === pluginConstants.common.boards || formFields.serviceType === pluginConstants.common.repos) {
    const subscriptionModalFields = {...subscriptionModalFields, eventType: {...subscriptionModalFields.eventType, isFieldDisabled: !formFields.project}, serviceType: {...subscriptionModalFields.serviceType, isFieldDisabled: !formFields.project}};
    subscriptionModalFields.eventType.optionsList = formFields.serviceType === pluginConstants.common.boards ? boardEventTypeOptions : repoEventTypeOptions;
    setSubscriptionModalFields(subscriptionModalFields);
}
```
You can do something like this.

webapp/src/components/dropdown/index.tsx
```diff
@@ -5,7 +5,7 @@ import './styles.scss';
 type DropdownProps = {
     value: string | null;
     placeholder: string;
-    onChange: (newValue: string, targetLabel?: string) => void;
+    onChange: (newValue: string, targetLabel?: string, selectedOption?: Record<string, string>) => void;
```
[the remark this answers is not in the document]
avas27JTG: no because the parent component where we will use this `Dropdown` will require to pass this `prop` as `LabelValuePair` only but sometimes we need to give a different object type whose type varies

manojmalik20 (approved):

PR #119, opened 2022-11-29
[MI-2405] Fix bug: Store small case string for organization name and …

raghavaggarwal2308:
…capital case for project name

### DIFF
--- a/server/plugin/api.go
+++ b/server/plugin/api.go
@@ -12,6 +12,8 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/mattermost/mattermost-server/v5/model"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 
 	"github.com/Brightscout/mattermost-plugin-azure-devops/server/constants"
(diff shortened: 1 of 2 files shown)


server/plugin/api.go
```diff
@@ -136,8 +138,8 @@ func (p *Plugin) handleLink(w http.ResponseWriter, r *http.Request) {
 	project := serializers.ProjectDetails{
 		MattermostUserID: mattermostUserID,
 		ProjectID:        response.ID,
-		ProjectName:      response.Name,
-		OrganizationName: body.Organization,
+		ProjectName:      cases.Title(language.Und).String(body.Project),
```
Kshitij-Katiyar, line 141: cases.Title(language.Und).String what does this do i am not familiar with this ?

Kshitij-Katiyar (approved):

server/plugin/api.go
raghavaggarwal2308: Converts the string to title case

avas27JTG (approved):

PR #123, opened 2022-12-01
[MI-2410] Updated subscription card to display run event filters and filter popover for pipelines

raghavaggarwal2308:


server/constants/constants.go
```diff
@@ -92,15 +92,16 @@ const (
...
-	FilterReleaseDefinitionID  = "releaseDefinitionId"
-	FilterReleaseEnvironmentID = "releaseEnvironmentId"
+	FilterCreatedByMe            = "me"
+	FilterCreatedByAnyone        = "anyone"
+	FilterAllChannels            = "all_channels"
+	FilterAll                    = "all"
+	FilterBoards                 = "boards"
+	FilterRepos                  = "repos"
+	FilterPipelines              = "pipelines"
+	EventTypeRelease             = "release"
```
ayusht2810, line 102: don't we need any prefix of `Filter` here?

webapp/src/components/card/subscription/index.tsx
```diff
@@ -97,6 +97,14 @@ const SubscriptionCard = ({handleDeleteSubscrption, subscriptionDetails: {channe
                                     {approvalStatusName && <Chip text={`Approval status is: ${approvalStatusName}`}/>}
                                     {approvalTypeName && <Chip text={`Approval type is: ${approvalTypeName}`}/>}
                                     {releaseStatusName && <Chip text={`Release status is: ${releaseStatusName}`}/>}
+                                    {runPipelineName && <Chip text={`Pipeline is: ${runPipelineName}`}/>}
+                                    {runEnvironment && <Chip text={`Environment is: ${runEnvironment}`}/>}
+                                    {runStage && <Chip text={`Stage is: ${runStage}`}/>}
+                                    {runStageId && <Chip text={`Stage is: ${runStageId}`}/>}
```
ayusht2810, line 103:
```suggestion
                                    {runStageId && <Chip text={`Stage Id is: ${runStageId}`}/>}
```

raghavaggarwal2308: The display name for this is actually "Stage" in the UI

server/constants/constants.go
raghavaggarwal2308: No, its not a filter. Separated it out to avoid the confusion

server/constants/constants.go
```diff
@@ -92,16 +92,17 @@ const (
...
+	FilterCreatedByAnyone      = "anyone"
+	FilterAllChannels          = "all_channels"
+	FilterAll                  = "all"
+	FilterBoards               = "boards"
+	FilterRepos                = "repos"
+	FilterPipelines            = "pipelines"
+	FilterReleaseDefinitionID  = "releaseDefinitionId"
+	FilterReleaseEnvironmentID = "releaseEnvironmentId"
+
+	EventOfTypeRelease = "release"
```
ayusht2810, line 105: i think the previous name was better

ayusht2810 (approved):

webapp/src/components/card/subscription/index.tsx
ayusht2810: One more chip is present above with the same data `Stage is:`. will the user be able to differentiate between them?

raghavaggarwal2308: Yes, they both can not exist at the same time

raghavaggarwal2308: they are for two different event types

server/constants/constants.go
raghavaggarwal2308: I changed this name because there is no eventType `release` but there are events which contain `release` in their name, that's what we are checking with this constant

webapp/src/components/card/subscription/index.tsx
avas27JTG, line 103:
```suggestion
                                    {(runStage || runStageId) && <Chip text={`Stage is: ${runStage || runStageId}`}/>}
```
see if this works, if both cannot exists together

webapp/src/components/card/subscription/index.tsx
```diff
@@ -97,6 +97,14 @@ const SubscriptionCard = ({handleDeleteSubscrption, subscriptionDetails: {channe
...
+                                    {runStageStateIdName && <Chip text={`Stage state is: ${runStageStateIdName}`}/>}
+                                    {runStageResultId && <Chip text={`Stage result is: ${runStageResultId}`}/>}
+                                    {runStateIdName && <Chip text={`State is: ${runStateIdName}`}/>}
```
avas27JTG, line 106: can be included in above too

raghavaggarwal2308: This one is `State`, above is `Stage`

avas27JTG (approved):

PR #127, opened 2022-12-02
Release v2.2.2

hanzei:
#### Summary
First release to the Marketplace

#### Open Tickets
- [ ] https://github.com/mattermost/mattermost-plugin-azure-devops/pull/128

#### Ticket Link
Part of #126

PR #128, opened 2022-12-02
Release to Marketplace

hanzei:
The sole purpose of this PR is to allow a review of the codebase of the plugin. It contains all changes that have been made since the repo was forked from https://github.com/mattermost/mattermost-plugin-starter-template.

The PR should not get merged. Once every party has approved the PR, #127 will include the required changes for a release.

Security review ticket: https://mattermost.atlassian.net/browse/SEC-2367.
Security review ticket for Phase 3: https://mattermost.atlassian.net/browse/SEC-3075.


srkgupta (changes requested):
Some vulnerabilities were observed and have been reported privately. Please check the JIRA ticket for more details.

server/plugin/client.go
```diff
@@ -0,0 +1,549 @@
...
+
+	case http.StatusNotFound:
+		return nil, resp.StatusCode, ErrNotFound
+	}
+
+	errResp := ErrorResponse{}
+	if err = json.Unmarshal(responseData, &errResp); err != nil {
+		return responseData, http.StatusInternalServerError, errors.WithMessagef(err, "status: %s", resp.Status)
+	}
+	return responseData, resp.StatusCode, fmt.Errorf("errorMessage %s", errResp.Message)
```
mickmister, line 525: For this type:

```go
type ErrorResponse struct {
	Message string `json:"message"`
}
```

As long as the response is valid JSON data, and it doesn't contain a `message` field that has a type other than `string`, this if block won't ever run. In other words, `message` is not necessarily defined when the marshal error does not occur. I'm thinking we should log the response code here since we potentially don't have an error message to use in this case.

mickmister: Bringing this up because there was an OAuth error @DHaussermann ran into when there was a configuration error for the app on Azure's side. The error message was blank so it read as: `errorMessage` with no message afterwards.

hanzei:
Closing as https://github.com/mattermost/mattermost-plugin-azure-devops/pull/153 has been merged.

PR #129, opened 2022-12-05
[MI-2428] Enhancement: Change the flow to list subscription: Check fo…

raghavaggarwal2308:
…r user id in case or user subscriptions and for project linked in cases of others subscriptions


server/plugin/utils.go
```diff
@@ -312,6 +312,16 @@ func (p *Plugin) GetSubscriptionsForAccessibleChannelsOrProjects(subscriptionLis
 		return nil, err
 	}
 
+	if createdBy == constants.FilterCreatedByMe {
+		var filteredSubscriptionList1 []*serializers.SubscriptionDetails
```
ayusht2810, line 316:
```suggestion
		var filteredSubscriptionList []*serializers.SubscriptionDetails
```

ayusht2810 (commented):
fix ci

server/plugin/utils.go
avas27JTG: and if there is issue with re-declaration you can use `filteredSubscriptionListForCurrentUser`

ayusht2810 (approved):

avas27JTG (approved):

PR #134, opened 2023-01-10
Migrate from CircleCI to Github actions

spirosoik:
#### Summary
We can use across all the repositories of plugins and apps the same reusable workflow which give us the ability to change only one file instead of multiple and apart from this the ability to use different versions as we prefer.

#### Ticket Link
Ticket: https://mattermost.atlassian.net/browse/CLD-4816


spirosoik:
@phoinixgrr @toninis good catch forgot to paste it. check again.

phoinixgrr (approved):
:shipit: 

A new era

toninis (approved):
🚀 That's the stuff

PR #144, opened 2023-02-24
[MI-2820]: Modified API endpoint to require a project to be provided

avas27JTG:
Modified API endpoint to require a project to be provided

### DIFF
--- a/server/plugin/api.go
+++ b/server/plugin/api.go
@@ -413,6 +413,21 @@ func (p *Plugin) handleGetSubscriptions(w http.ResponseWriter, r *http.Request) 
 		return
 	}
 
+	projectList, err := p.Store.GetAllProjects(mattermostUserID)
+	if err != nil {
+		p.API.LogError(constants.ErrorFetchProjectList, "Error", err.Error())
+		p.handleError(w, r, &serializers.Error{Code: http.StatusInternalServerError, Message: err.Error()})
+		return
+	}
+
+	organization := pathParams[constants.PathParamOrganization]
+	project := pathParams[constants.PathParamProject]
+	if _, isProjectLinked := p.IsProjectLinked(projectList, serializers.ProjectDetails{OrganizationName: strings.ToLower(organization), ProjectName: cases.Title(language.Und).String(project)}); !isProjectLinked {
+		p.API.LogError(fmt.Sprintf("Project %s is not linked", project))
+		p.handleError(w, r, &serializers.Error{Code: http.StatusBadRequest, Message: "requested project is not linked"})
+		return
+	}
+
 	var subscriptionList []*serializers.SubscriptionDetails
 	var subscriptionErr error
 	createdBy := r.URL.Query().Get(constants.QueryParamCreatedBy)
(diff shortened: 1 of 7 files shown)


server/plugin/api.go
```diff
@@ -413,6 +413,21 @@ func (p *Plugin) handleGetSubscriptions(w http.ResponseWriter, r *http.Request)
...
+	projectList, err := p.Store.GetAllProjects(mattermostUserID)
+	if err != nil {
+		p.API.LogError(constants.ErrorFetchProjectList, "Error", err.Error())
+		p.handleError(w, r, &serializers.Error{Code: http.StatusInternalServerError, Message: err.Error()})
+		return
+	}
+
+	organization := pathParams[constants.PathParamOrganization]
+	project := pathParams[constants.PathParamProject]
+	if _, isProjectLinked := p.IsProjectLinked(projectList, serializers.ProjectDetails{OrganizationName: strings.ToLower(organization), ProjectName: cases.Title(language.Und).String(project)}); !isProjectLinked {
```
mickmister, line 425: There's a lot going on in this line. Maybe break this into local variables first?

server/plugin/api.go
```diff
@@ -413,6 +413,21 @@ func (p *Plugin) handleGetSubscriptions(w http.ResponseWriter, r *http.Request)
...
+		p.API.LogError(fmt.Sprintf("Project %s is not linked", project))
```
mickmister, line 426: I think we should use `LogWarn` since there isn't an actionable step for the admin to fix this. Please keep this is mind for other features/fixes implemented.

server/plugin/api.go
```diff
@@ -413,6 +413,21 @@ func (p *Plugin) handleGetSubscriptions(w http.ResponseWriter, r *http.Request)
...
 
+	projectList, err := p.Store.GetAllProjects(mattermostUserID)
+	if err != nil {
+		p.API.LogError(constants.ErrorFetchProjectList, "Error", err.Error())
+		p.handleError(w, r, &serializers.Error{Code: http.StatusInternalServerError, Message: err.Error()})
+		return
+	}
+
+	organization := pathParams[constants.PathParamOrganization]
+	project := pathParams[constants.PathParamProject]
```
mickmister, line 424: We could "return early" if `project` is blank right? Before we interact with the store

hanzei:
@avas27JTG Could you please rebase against `master` and change the target branch?

server/plugin/api.go
avas27JTG: Yes, we have made `organization` and `project` path params so, if they are empty this API will not be called as they will not reach here and get `404`

server/plugin/api.go
avas27JTG: Sure, fixed

server/plugin/api.go
```diff
@@ -413,6 +413,26 @@ func (p *Plugin) handleGetSubscriptions(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
+	projectList, err := p.Store.GetAllProjects(mattermostUserID)
+	if err != nil {
+		p.API.LogError(constants.ErrorFetchProjectList, "Error", err.Error())
```
mickmister, line 418: Same comment about using `LogWarn` instead here

server/plugin/api.go
```diff
@@ -432,7 +452,6 @@ func (p *Plugin) handleGetSubscriptions(w http.ResponseWriter, r *http.Request)
 	channelID := r.URL.Query().Get(constants.QueryParamChannelID)
 	serviceType := r.URL.Query().Get(constants.QueryParamServiceType)
 	eventType := r.URL.Query().Get(constants.QueryParamEventType)
-	project := r.URL.Query().Get(constants.QueryParamProject)
 	if project != "" {
```
mickmister, line 455: What's the intended functionality when `project ` is an empty string? The code reads a bit weird because there is a large block for handling when it's not empty, but then the function ends after that block

avas27JTG: The code will reach here only when the `project` is not empty and it's valid i.e. linked/exists on the plugin
and in that case, we are returning the array of subscriptions

Removed this check as it's no longer needed now as we have made it required now

mickmister (approved):

PR #147, opened 2023-03-13
[MM-51250]: Added max limit for reading responses

avas27JTG:
See MM-51250

### DIFF
--- a/server/plugin/client_test.go
+++ b/server/plugin/client_test.go
@@ -632,7 +632,7 @@ func TestMakeHTTPRequest(t *testing.T) {
 		},
 		{
 			description:                       "MakeHTTPRequest: large response body",
-			maxBytesSizeForClientResponseBody: 10000000,
+			maxBytesSizeForClientResponseBody: constants.MaxBytesSizeForReadingResponseBody + 1,
 		},
 	} {
 		t.Run(testCase.description, func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


server/plugin/client.go
```diff
@@ -497,6 +497,10 @@ func (c *client) makeHTTPRequest(req *http.Request, contentType string, out inte
 	}
 	defer resp.Body.Close()
 
+	// Limit reading the response bodies to 1 MB or 1000 KB
+	// This is ideal for the responses returned by Azure DevOps APIs used here
+	resp.Body = http.MaxBytesReader(nil, resp.Body, constants.MaxBytesSizeForReadingResponseBody)
```
mickmister, line 502: It seems off to reassign `resp.Body` here. Thoughts on making this its own var?

avas27JTG: `resp.Body` was not used further after this so reassigned to it, but yes having a separate variable for it will make it more clear

mickmister (approved):

jupenur (approved):
A test case would be nice to have. Otherwise LGTM.

srkgupta (approved):

m1lt0n (approved):

avas27JTG:
added m1lt0n as non-blocking reviewer

avas27JTG:
> A test case would be nice to have. Otherwise LGTM.

Added a test case.

PR #148, opened 2023-03-15
[MM-51253][MM-51275]: Added check for full name and refactored error logs.

avas27JTG:
See [MM-51253] and [MM-51275]


server/plugin/oAuth.go
```diff
@@ -126,9 +126,6 @@ func (p *Plugin) GenerateOAuthToken(code, state string) error {
 	mattermostUserID := strings.Split(state, "_")[1]
 
 	if err := p.Store.VerifyOAuthState(mattermostUserID, state); err != nil {
-		if _, DMErr := p.DM(mattermostUserID, constants.GenericErrorMessage, false); DMErr != nil {
```
mickmister, line 129: Can we also make sure every endpoint intended to be used by a MM user is checked for authentication?

mickmister (changes requested):

server/plugin/oAuth.go
avas27JTG: Yeah, we have a check for authentication on all the plugin APIs where we are checking for MM user-id in the header.
Also, `OAuthComplete` API is used as a redirection URL by the Azure DevOps oAuth app which sends the `state` query param that we are validating here as authentication.

m1lt0n (approved):

server/plugin/oAuth.go
mickmister: > Also, `OAuthComplete` API is used as a redirection URL by the Azure DevOps oAuth app which sends the `state` query param that we are validating here as authentication.

The redirect is in the browser only though. Any request hitting MM in the process will be authenticated with a MM user token. The Azure server never sends a request to the MM server in this process.

mickmister: We should also check that the authenticated user's id matches the user id on line 126

avas27JTG: Made the changes

mickmister (approved):

PR #149, opened 2023-03-15
[MM-51247][MM-51248]: Added validations for path inputs and fixed typo.

avas27JTG:
See [MM-51247] and [MM-51248]


server/plugin/client.go
```diff
@@ -531,7 +532,8 @@ func (c *client) makeHTTPRequestWithAccessToken(basePath, path, method, accessTo
 		return nil, http.StatusInternalServerError, err
 	}
 
-	req, err := http.NewRequest(method, URL, nil)
+	sanitizedURL := sanitize.URL(URL)
+	req, err := http.NewRequest(method, sanitizedURL, nil)
```
mickmister, line 536: As noted in the ticket:

> This also applies to all other API client methods and should be fixed consistently everywhere

This fix needs to be applied for all outgoing requests, but this function `makeHTTPRequestWithAccessToken` is only called by `GetUserProfile`.

mickmister (changes requested):

server/plugin/client.go
jupenur: This isn't a fix at all. `mrz1836/go-sanitize` might be the first hit when you google for "golang sanitize url" but it has nothing to do with what we're trying to do here. I would not be comfortable shipping that library even if it was attempting to do the correct thing, which it's not.

Also from the ticket:

> The `organization`, `*ID` and `projectName` variables are fully user controlled and must be sanitized in case of `../` injections.

It's the input variables that need to be sanitized before they get used as URL path components, not the completed URL. The thing to look for is also named right there. This should not require any external dependencies. There are multiple possible approaches, but the stdlib `strings` and `path` packages might be helpful.

jupenur (changes requested):
The sanitization approach is completely misguided. Also please include tests! Even a simple test would show you that this fix is broken.

server/plugin/client.go
m1lt0n: +1 on no need for an external dependency. The fix should be applied possibly only to the `path` part of the input and also be applied to all API client methods.

m1lt0n (changes requested):

server/plugin/client.go
avas27JTG: We removed the external package and used `path.Clean` to sanitize URL paths in all the needed client methods, please let us know if this looks good now

server/plugin/client.go
```diff
@@ -56,9 +57,11 @@ func (c *client) GenerateOAuthToken(encodedFormValues url.Values) (*serializers.
 }
 
 func (c *client) GetUserProfile(id, accessToken string) (*serializers.UserProfile, int, error) {
-	var userProfile *serializers.UserProfile
+	userProfilePath := fmt.Sprintf(constants.PathUserProfile, id)
+	sanitizedPath := path.Clean(userProfilePath)
```
m1lt0n, line 61: if `userProfilePath` does not start with a forward slash, `Clean` e.g. on "../x/y/z" will be "../x/y/z", ie unchanged. A unit test would reveal that.

server/plugin/client.go
```diff
@@ -100,7 +104,7 @@ func (c *client) CreateTask(body *serializers.CreateTaskRequestPayload, mattermo
 	}
 
 	var task *serializers.TaskValue
-	_, statusCode, err := c.CallPatchJSON(c.plugin.getConfiguration().AzureDevopsAPIBaseURL, taskURL, http.MethodPost, mattermostUserID, &payload, &task, nil)
+	_, statusCode, err := c.CallPatchJSON(c.plugin.getConfiguration().AzureDevopsAPIBaseURL, sanitizedPath, http.MethodPost, mattermostUserID, &payload, &task, nil)
```
m1lt0n, line 107: (suggestion) since the same operation is repeated in multiple places, and all `Call...Json` use client's `Call` method, I suggest performing any transformations for path in `parsePath` that's used by `Call` and also directly by `makeHTTPRequestWithAccessToken` which doesn't use `Call`

m1lt0n (changes requested):

server/plugin/utils_test.go
```diff
@@ -356,6 +356,41 @@ func TestGetPluginURL(t *testing.T) {
...
+		},
+		{
+			description:  "SanitizeURLPath: invalid variation 2",
+			actualPath:   "../dummy_org/../dummy_project/..",
+			expectedPath: "/dummy_org/dummy_project",
+		},
+		{
+			description:  "SanitizeURLPath: invalid variation 3",
+			actualPath:   "../../dummy_org/dummy_project",
+			expectedPath: "/dummy_org/dummy_project",
```
m1lt0n, line 384: (question) is the path always a non-root one? (ie `abd/def` compared to `/abcd/ef'`)? Because if this is the case, would the extra slash break API calls? Did a manual test take place in the context of this change?

jupenur (changes requested):
This is unfortunate and is going to complicate things a bit more, but I just did some testing, and... this call returns a valid API response:

```
curl -v --path-as-is -X GET 'https://dev.azure.com/juhonurminen/_apis/potato/%2e%2e/hooks/subscriptions?api-version=6.0' -H 'Authorization: Basic <redacted>'
```

That means we need to explicitly handle URL encoding here as well, because Microsoft is being dumb about the way they do routing. `%2e%2e` needs to be treated the same way as `..`, as do all combinations of the two.

server/plugin/utils_test.go
avas27JTG: there would not be extra slash as we have used `path.Clean` method which removes any extra slash left after removing `..` or `%2e%2e`
Yes, have tested manually Azure Devops APIs are working fine

server/plugin/utils.go
```diff
@@ -493,3 +494,9 @@ func (p *Plugin) CheckValidChannelForSubscription(channelID, userID string) (int
 
 	return 0, nil
 }
+
+func (p *Plugin) SanitizeURLPath(urlPath string) string {
+	parsedPath := strings.ReplaceAll(strings.ReplaceAll(urlPath, "..", ""), "%2e%2e", "")
```
jupenur, line 499: This covers the simple `..` and `%2e%2e` cases, but by "combinations of the two" I meant we also need to cover things like `.%2e`, `%2e.`, `.%2e%2e.`, `.%2e%2e%2e` and so forth.

avas27JTG: tried with regex to check for different combinations of `.` and `%2e` to eliminate path traversal patterns like `../`, `./`, `/../..`, etc
also if there is any handy doc can you please let me know what are the different harmful characters and their combination that we should prevent putting in a URL

avas27JTG: also Azure Devops allow putting dots in between other characters in their project name like `dummy..1..2..3..project` so, handled this also

jupenur: Since you're now matching `../` instead of just `..` like you previously did, the encoded form of the forward slash also becomes relevant, i.e `..%2f` and all other variations containing `%2f` need to be taken care of.

> also if there is any handy doc, can you please let me know?

There is no good documentation that you could directly apply because every scenario is a little different and these ones where the path traversal affects a server-side REST API call is a somewhat poorly recognized one in particular. You can find some general background info on [Wikipedia](https://en.wikipedia.org/wiki/Directory_traversal_attack) or the [OWASP Testing Guide](https://github.com/OWASP/wstg/blob/master/document/4-Web_Application_Security_Testing/05-Authorization_Testing/01-Testing_Directory_Traversal_File_Include.md), though, but don't take it at face value.

As you've noticed, the regexp/replace approach, especially applied to the completed URL instead of the inputs, gets very complicated very quickly and it's easy to make subtle mistakes. That's why my [original recommendation](https://github.com/mattermost/mattermost-plugin-azure-devops/pull/149#discussion_r1137684782) was to sanitize the inputs instead *before* incorporating them in the API path. It allows you to make better use of the stdlib `path.Clean` and `net/url.PathUnescape` functions and ditch the regexes.

Although I think @m1lt0n was the one to [suggest the current approach](https://github.com/mattermost/mattermost-plugin-azure-devops/pull/149#discussion_r1138527959), so maybe he has some other insights.

server/plugin/utils_test.go
```diff
@@ -356,6 +356,81 @@ func TestGetPluginURL(t *testing.T) {
...
+			expectedPath: "/dummy_org/dummy..1..2..3..project/dummy.project/dummy....project",
+		},
+		{
+			description:  "SanitizeURLPath: invalid variation 1",
+			actualPath:   "/../dummy_org/dummy_project",
+			expectedPath: "/dummy_org/dummy_project",
+		},
+		{
+			description:  "SanitizeURLPath: invalid variation 2",
+			actualPath:   "../dummy_org/../dummy_project/..",
```
m1lt0n, line 383: Shouldn't the expected value of this be `dummy_org/dummy_project` instead of starting with a forward slash?

Regarding `SanitizeURLPath`, I think the logic should be:

1. unescapedPath (replace %2e with . and %2f with /)
2. if path starts with / , a simple `path.Clean(actualPath)` should be enough
3. if path does not start with /, doing `strings.TrimPrefix("/", path.Join("/", actualPath))` should work properly

Example inputs and outputs ( opinion on this @jupenur ):

```
/dummy_org/dummy_project -> /dummy_org/dummy_project
/dummy_org/dummy..1..2..3..project/dummy.project/dummy....project -> /dummy_org/dummy..1..2..3..project/dummy.project/dummy....project
/../dummy_org/dummy_project -> /dummy_org/dummy_project
../dummy_org/../dummy_project/.. -> ""
../../dummy_org/../../dummy_project -> dummy_project
/dummy_org/../dummy_project ->  /dummy_project
/dummy_org/%2e%2e/dummy_project -> /dummy_project
/dummy_org/.%2e/dummy_project -> /dummy_project
/dummy_org/%2e./dummy_project -> /dummy_project
/dummy_org/.%2e/%2e./dummy_project -> /dummy_project
/dummy_org/.%2e%2e./dummy_project/.. -> /dummy_org/....
%2e%2e/dummy_org/.%2e%2e./dummy_project/.%2e.%2e/ -> dummy_org/..../dummy_project/....
```

jupenur: > 2. if path starts with / , a simple `path.Clean(actualPath)` should be enough

No. The path here is the actual full REST API path with inputs already baked in. Just calling `Clean` on it doesn't fix anything -- it just normalizes the path, but Azure is already doing that on their end before routing, so there's no impact there.

The thing we're trying to prevent here (which apparently is not clear) is this:

1. There is an API call to `/some/path/with/${inputs}`
2. The user sets `inputs` to `../../other/path`
3. We issue an API call to `/some/path/with/../../other/path`
4. Azure treats the call as a call to `/some/other/path`

Calling `Clean` does not solve this, because with `Clean` the path `/some/path/with/../../other/path` just becomes `/some/other/path`. There is no "escaping the root" because we're already dealing with the completed path. The root is what it is, and Azure takes care of that, but the input should not change the API endpoint being called.

avas27JTG: Adding to what @jupenur mentioned instead of checking the entire URL we should check and sanitize individual inputs that are going to be part of the URL, this way we can allow having only those characters in such inputs which are supported by Azure DevOps APIs removing all non-allowed chars like `@, /`, etc and also removing suspicious path traversal patterns formed by `.` and `%2e` like `../`, `/..` `/../..`, `/%2e%2e`, etc

let me know if this seems good to you and we proceed with it @jupenur @m1lt0n

m1lt0n (approved):

jupenur (changes requested):
One last thing I failed to notice earlier: Azure also treats `\` as a path separator, so we also need to deal with that in `SanitizeURLPaths`. Currently it's possible to still pass a path like `x%5c..%5c..%5cother%5capi` as a project name.

The sanitization function looks much more complicated than what I originally envisioned, but if @m1lt0n and @mickmister are ok with it, I am too, since after some final minor changes it seems like it should work correctly. I'm also not a huge fan of the regexes, both because of readability, and because we're recompiling them on every invocation, but again it's not my place to comment on that since there is no security concern there.

Getting there!

server/plugin/client.go
```diff
@@ -272,20 +301,23 @@ func (c *client) DeleteSubscription(organization, subscriptionID, mattermostUser
...
 	baseURL = strings.Replace(baseURL, "://", "://vsrm.", 1)
-	_, statusCode, err := c.CallJSON(baseURL, pipelineApproveRequestURL, http.MethodPatch, mattermostUserID, &pipelineApproveRequestPayload, nil, nil)
+	_, statusCode, err := c.CallJSON(baseURL, updatePipelineApproveRequestPath, http.MethodPatch, mattermostUserID, &pipelineApproveRequestPayload, nil, nil)
 
 	return statusCode, err
 }
 
 func (c *client) UpdatePipelineRunApprovalRequest(pipelineApproveRequestPayload []*serializers.PipelineApproveRequest, organization, projectID, mattermostUserID string) (*serializers.PipelineRunApproveResponse, int, error) {
-	pipelineApproveRunRequestURL := fmt.Sprintf(constants.PipelineRunApproveRequest, organization, projectID)
+	updatePipelineApproveRunRequestPath := fmt.Sprintf(constants.PipelineRunApproveRequest, organization, projectID)
```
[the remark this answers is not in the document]
avas27JTG: Added

avas27JTG:
> 

Added the check for `%5c` or `\` also

jupenur (approved):

mickmister (approved):

mickmister:
@dhaussermann @m1lt0n Is the plan to have internal QA review on a PR like this?

m1lt0n:
@DHaussermann based on the planning discussion today then, we can merge this right? cc: @mickmister

PR #150, opened 2023-03-15
[MI-51272]: Refactored and removed unused code.

avas27JTG:
See [MI-51272]


server/plugin/client.go
```diff
@@ -207,12 +207,7 @@ var publisherID = map[string]string{
 func (c *client) CreateSubscription(body *serializers.CreateSubscriptionRequestPayload, project *serializers.ProjectDetails, channelID, pluginURL, mattermostUserID string) (*serializers.SubscriptionValue, int, error) {
 	subscriptionURL := fmt.Sprintf(constants.CreateSubscription, body.Organization)
 
-	encryptedWebhookSecret, err := c.plugin.Encrypt([]byte(c.plugin.getConfiguration().WebhookSecret), []byte(c.plugin.getConfiguration().EncryptionSecret))
-	if err != nil {
-		return nil, http.StatusInternalServerError, errors.Wrap(err, "failed to encrypt webhook secret")
-	}
-	encodedWebhookSecret := c.plugin.Encode(encryptedWebhookSecret)
-
+	encodedWebhookSecret := c.plugin.Encode([]byte(c.plugin.getConfiguration().WebhookSecret))
```
mickmister, line 210: Just so I understand, is this base64 encoding required?

m1lt0n: +1 on @mickmister comment, is the encoding/decoding needed?

avas27JTG: we are generating `webhookSecret` using `generated` type in the plugin's setting which adds some characters that are not accepted by Azure DevOps APIs so we are encoding them before sending

jupenur: Right. [It's supposed to be the URL-friendly variant of Base64](https://github.com/mattermost/mattermost-webapp/blob/780bbe87b2b40de0cbb6359648a90aae672f5ff5/components/admin_console/schema_admin_settings.jsx#L673-L675), but there *is* a bug there.

I've filed [MM-51451](https://mattermost.atlassian.net/browse/MM-51451).

Regardless of the bug, a better solution here would be coercing the value back to the standard base64-encoding rather than re-encoding it:

```go
strings.ReplaceAll(strings.ReplaceAll(c.plugin.getConfiguration().WebhookSecret, "_", "/"), "-", "+")
```

avas27JTG: Sure it's better to coerce

avas27JTG: Some of the characters are not valid as query-param like `+` is replaced with ` `(space) when parsing the query-param so to avoid such edge cases have escaped the string so it can be safely placed inside a URL query using method `url.QueryEscape` which is working correctly for all such cases.

Example of some generated values that creates issue `gc6y-Uayt+aufBbwZ73NCCsAYZH62aiP`, `x3NI6UxUIxifIaUkQqBpU5b++iXoYKKF`

jupenur: `url.QueryEscape` does not escape `-` which you previously mentioned as problematic. And `+` *is* part of the standard Base64 charset, which is what you were originally using but now that doesn't work either?

Are you testing this properly? If you know which characters the API accepts and which it doesn't, can you add an unit test for that?

Also is there any reason to wrap the `url.QueryEscape` in another function call? Seems like it's just unnecessarily complicating things.

avas27JTG: There is no written doc for supported chars but I tried for multiple chars seems like API doesn't have a problem with accepting them, it was the query-param parsing that was giving errors while reading the secret returned from the API, if we pass escaped string everything seems to be working now, have tested this manually.

Regarding the test case, it seems like all encoded characters are sent and received properly, let me know if we still need to add a separate test case for it as I could not get any char generated randomly in the plugin's generated field causing the issue now so, if we add test case it will be mostly valid ones.

m1lt0n (approved):

jupenur (approved):

mickmister (approved):

m1lt0n:
@DHaussermann based on the planning discussion today then, we can merge this right? cc: @mickmister

PR #152, opened 2023-04-03
[MM-51404]: Added logic to generate unique webhook secret for subscriptions

avas27JTG:
Added logic to generate unique webhook secret for subscriptions


server/store/subscriptions.go
```diff
@@ -198,3 +203,30 @@ func SubscriptionListFromJSON(bytes []byte) (*SubscriptionList, error) {
...
+}
+
+func (s *Store) GetSubscriptionChannelID(subscriptionID string) (*SubscriptionWebhookSecretAndChannelMap, error) {
+	var storedWebhookSecret SubscriptionWebhookSecretAndChannelMap
+	if err := s.LoadJSON(subscriptionID, &storedWebhookSecret); err != nil {
+		return nil, err
+	}
+
+	return &storedWebhookSecret, nil
+}
```
mickmister, line 224: It seems off that the method `GetSubscriptionChannelID` is returning something called a webhook secret. The returned value is not a string either, which it seems like `secret` would be a string

server/store/subscriptions.go
```diff
@@ -198,3 +203,30 @@ func SubscriptionListFromJSON(bytes []byte) (*SubscriptionList, error) {
 	}
 	return subscriptionList, nil
 }
+
+func (s *Store) StoreSubscriptionChannelID(subscriptionID, webhookSecret, channelID string) error {
+	if err := s.StoreJSON(subscriptionID, SubscriptionWebhookSecretAndChannelMap{
+		webhookSecret: channelID,
```
mickmister, line 209: This reads a little weird, because it's not obvious that `StoreSubscriptionChannelID` is literally a map, and not a struct with a field named `webhookSecret`

mickmister (approved):
Just have a few non-blocking comments. LGTM :+1:

m1lt0n (approved):

server/store/subscriptions.go
avas27JTG: fixed

server/store/subscriptions.go
avas27JTG: fixed

esarafianou (approved):
LGTM

